PR #45, opened 2014-11-20
ARQ-1888 Type mapping for IE capabilites

yurloc:
I have added all Boolean and Integer IE specific and common capabilities. This will prevent IE driver failure and later, when the fix for [Issue 8160](https://code.google.com/p/selenium/issues/detail?id=8160) is released, this type mapping will prevent IE driver from ignoring wrongly typed capabilities.


kpiwko:
Landed in 2.x stream, will be backported to 1.x stream - see https://issues.jboss.org/browse/ARQ-1906

kpiwko:
Thanks so far!

MatousJobanek:
Closing as already merged

PR #49, opened 2015-01-21
Add precedence for BeforeClass to make sure we run before deployment.

TadeasKriz:
This allows containers to put metadata to deployment-scoped drones.


kpiwko:
@smiklosovic this should definitely go to next alpha

kpiwko:
Landed in https://github.com/arquillian/arquillian-extension-drone/commit/6baa91caba312103eafe252cbc2ec092cf842671

Thanks!

PR #50, opened 2015-01-25
Alpha3 release

smiklosovic:
@kpiwko it is already in central


smiklosovic:
@kpiwko I do not have push rights for Drone repository.

PR #54, opened 2015-10-16
ARQ-1692 Factories are registered based on WebDriver's presence on class path

MatousJobanek:
It is not very nice solution, but probably the only one without modifying arquillian-core behaviour (specifically this https://github.com/arquillian/arquillian-core/blob/master/core/impl-base/src/main/java/org/jboss/arquillian/core/impl/loadable/ServiceRegistryLoader.java#L103-L108)
 
It basically check whether the required WebDriver of some particular factory is on class path. If not, the warning is printed and the factory is not registered.

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/DroneWebDriverExtension.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/DroneWebDriverExtension.java
@@ -94,7 +94,7 @@ public class DroneWebDriverExtension implements LoadableExtension {
     }
 
     private <T extends Configurator & Instantiator & Destructor> void registerFactoryService(
-        ExtensionBuilder builder, Class<? extends T> factory, String expectedDriver) {
+        ExtensionBuilder builder, Class<T> factory, String expectedDriver) {
 
         try {
             Class.forName(expectedDriver, false, this.getClass().getClassLoader());
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/DroneWebDriverExtension.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/DroneWebDriverExtension.java
@@ -16,6 +16,9 @@
  */
 package org.jboss.arquillian.drone.webdriver;
 
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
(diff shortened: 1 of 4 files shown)


kpiwko:
:+1:  I like this solution!

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/DroneWebDriverExtension.java
```diff
@@ -101,4 +92,20 @@ public void register(ExtensionBuilder builder) {
         builder.service(DroneInstanceEnhancer.class, AugmentingEnhancer.class);
         builder.observer(WindowResizer.class);
     }
+
+    private <T extends Configurator & Instantiator & Destructor> void registerFactoryService(
+        ExtensionBuilder builder, Class<? extends T> factory, String expectedDriver) {
```
TadeasKriz: Is `? extends T` necessary when you use the factory parameter as the constraint itself so it can never be anything but `Class<T>`?

TadeasKriz:
Does this solution work properly? It seems to me that it would not allow you to register `ChromeDriverFactory.class`, because it imports the `ChromeDriver` and I remember there were some `ClassDefNotFound` exceptions when using the class that imports classes not on the classpath.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/DroneWebDriverExtension.java
MatousJobanek: Nice catch!!! It is not necessary here - I'll change it. Thanks @TadeasKriz :-)

MatousJobanek:
Hi @TadeasKriz 
to be honest, I don't know what you exactly mean. When the `ChromeDriver class` was not on the classpath, there were thrown the kind of exception you mentioned during the instantiation of `ChromeDriverFactory.class`. And this is what this commit fixes - it checks whether the required webdriver class is on classpath and if not, the factory is not registered so as not to be instantiated later. 
I attached test to prove that it should be working properly.

TadeasKriz:
Nice, thanks for the test.

MatousJobanek:
Merged https://github.com/arquillian/arquillian-extension-drone/commit/ff8936b14927de21805d15c63083c95c9b3efb20

PR #57, opened 2016-01-12
ARQ-2006 creation of deployment-scoped drone points was postponed to AfterDeploy

MatousJobanek:
Hi @TadeasKriz @kpiwko may I ask you for a review?
@smiklosovic could you take a look if it is acceptable by Droidium? I ran the tests and after small modifications it passed, so I hope that everything will be OK.

Thanks :-)

---
current workflow is:
BeforeClass - deployment-scoped drone points are only registered
Enrichment phase - depends if the deployment is done or not - if yes,
then the drone points are injected. If not, then the test class instance
is registered
AfterDeploy - If it is before the enrichment phase then the corresponding
drone points are prepared. If it is after, then they are also injected.

What is required?
- A deployment associated to drone points used as a method parameter has
  to be deployed before the method is launched.

What is not covered?
- multiple manual deployment and undeployment within one test method


MatousJobanek:
The tests will be attached when finished - currently working on them.

smiklosovic:
Sorry I do not have time for this now ...

MatousJobanek:
tests added
@smiklosovic never mind - I tested it, so I hope that everything be OK

drone-impl/src/main/java/org/jboss/arquillian/drone/impl/DeploymentDronePointsRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+import org.jboss.arquillian.container.test.api.OperateOnDeployment;
+import org.jboss.arquillian.drone.spi.DronePoint;
+
+/**
+ * This registry keeps all {@link DronePoint}s with the {@link DronePoint.Lifecycle.DEPLOYMENT} scope and the reference
+ * to the instance of the test class the DronePoint is declared in
+ *
+ * @author <a href="mailto:mjobanek@redhat.com">Matous Jobanek</a>
+ */
+public class DeploymentDronePointsRegistry {
```
kpiwko: Is there a separate registry per Drone point lifecycle? If not, I'd rather think of having just on registry class.

MatousJobanek: well... no, there is no registry per Drone point lifecycle. 
Do you mean that you would welcome a registry for all the Drone points? There is something similar - DroneContext - it keeps all the Drone points, however, some additional changes would be needed (related to the inner implementation) and this class is available there in the SPI. So, I didn't want to change the SPI because of these changes and didn't want to make the information available through the SPI. The reason why I created a new separate registry is because this will be needed only in one case - when a deployment-scoped drone point is used - so mostly the registry won't be needed and won't be created. Not sure if it is good approach or not - that's why I'm asking for the review :-).

PR #62, opened 2016-03-16
Arquillian Drone BrowserStack extension

MatousJobanek:


drone-bom/pom.xml
```diff
@@ -80,6 +80,13 @@
                 <version>${project.version}</version>
             </dependency>
 
+            <!-- Arquillian Drone extension -->
```
smiklosovic: Arquillian Drone BrowserStack extension

extension/arquillian-drone-browserstack-extension/README.adoc
```diff
@@ -0,0 +1,138 @@
+= Arquillian Drone BrowserStack extension
+This is an Arquillian extension that enables you to use the BrowserStack as a Drone WebDriver there in your functional tests.
+
+=== How to use it:
+
+1):: First of all, you need to have this extension on your project's classpath. Eg. as a maven dependency:
```
smiklosovic: Maven

extension/arquillian-drone-browserstack-extension/README.adoc
```diff
@@ -0,0 +1,138 @@
...
+2):: To use the BrowserStack driver you need to specify *browserstack* as a *browser* property:
+[source,xml]
+----
+    <extension qualifier="webdriver">
+        <property name="browser">browserstack</property>
+        ...
+    </extension>
+----
+
+3):: You can choose if you want to specify your credentials to access your account or to provide the URL directly with the credentials included.
```
smiklosovic: I would add also the possibility to specify this username and access.key on the command line via some environment property. In case you run this somewhere, you can not commit your credentials to repository. You just want to have it parametrized during runtime.

extension/arquillian-drone-browserstack-extension/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+        <!-- Drone dependencies -->
+        <dependency>
+            <groupId>org.jboss.arquillian.extension</groupId>
+            <artifactId>arquillian-drone-webdriver</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.arquillian.spacelift</groupId>
+            <artifactId>arquillian-spacelift</artifactId>
+            <version>1.0.0.Alpha7</version>
```
smiklosovic: there is already alpha9

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
```diff
@@ -0,0 +1,257 @@
...
+    private BrowserStackLocalRunner() {
+    }
+
+    /**
+     * Returns an instance of BrowserStackLocalRunner. If there has been already created, returns this one, otherwise
+     * creates and returns a new one - behaves like singleton
+     *
+     * @return An instance of BrowserStackLocalRunner
+     */
+    public static BrowserStackLocalRunner createBrowserStackLocalInstance() {
```
smiklosovic: you could just do line 47 public static final BrowserStackLocalRunner broser = new ... with the same effect.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * Indirectly runs BrowserStackLocal binary. In case that the binary has been already run, then does nothing.
+     *
+     * @param accessKey      An accessKey the binary should be ran with
+     * @param additionalArgs additional arguments
+     * @param localBinary    Path to a local binary of the BrowserStackLocal. If none, then it will be downloaded.
+     */
+    public void runBrowserStackLocal(String accessKey, String additionalArgs, String localBinary) {
+        if (process != null) {
+            return;
```
smiklosovic: maybe log why you return?

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
```diff
@@ -0,0 +1,257 @@
...
+                                                    + "Supported platforms are windows, linux and macosx."
+                                                    + "Your platform has been detected as "
+                                                    + PlatformUtils.platform().os().toString().toLowerCase() + ""
+                                                    + "from the the system property 'os.name' => '" + PlatformUtils.OS
+                                                    + "'.");
+
+        }
+    }
+
+    private boolean isEmpty(String object) {
```
smiklosovic: name of that method is missleading, what about "isNullOrEmpty"?

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
```diff
@@ -0,0 +1,257 @@
...
+     */
+    public void runBrowserStackLocal(String accessKey, String additionalArgs, String localBinary) {
+        if (process != null) {
+            return;
+        }
+        if (isEmpty(localBinary)) {
+            if (!browserStackLocalFile.exists()) {
+                prepareBrowserStackLocal();
+            }
+            runTheBinary(browserStackLocalFile, accessKey, additionalArgs);
```
smiklosovic: I would just personally never call a method "runTheBinary" ...

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
```diff
@@ -0,0 +1,257 @@
...
+
+        log.info("downloading zip file from: " + url + " to " + browserStackLocalZipFile.getPath());
+        Spacelift.task(DownloadTool.class).from(url).to(browserStackLocalZipFile.getPath()).execute().await();
+
+        log.info("extracting zip file: " + browserStackLocalZipFile + " to " + browserStackLocalDirectory.getPath());
+        Spacelift.task(browserStackLocalZipFile, UnzipTool.class).toDir(browserStackLocalDirectory.getPath()).execute()
+            .await();
+
+        log.info("marking binary file: " + browserStackLocalFile.getPath() + " as executable");
+        browserStackLocalFile.setExecutable(true);
```
smiklosovic: this could potentially throw exception you do not catch nor log anywhere

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
```diff
@@ -0,0 +1,257 @@
...
+        String url = basicUrl + platformBinaryNameUrl;
+
+        log.info("Creating directory: " + browserStackLocalDirectory);
+        browserStackLocalDirectory.mkdir();
+
+        log.info("downloading zip file from: " + url + " to " + browserStackLocalZipFile.getPath());
+        Spacelift.task(DownloadTool.class).from(url).to(browserStackLocalZipFile.getPath()).execute().await();
+
+        log.info("extracting zip file: " + browserStackLocalZipFile + " to " + browserStackLocalDirectory.getPath());
+        Spacelift.task(browserStackLocalZipFile, UnzipTool.class).toDir(browserStackLocalDirectory.getPath()).execute()
```
smiklosovic: just a hint, these commands tends to be too long and harder to read, i prefer something like this:

`
Spacelift.task(browserStackLocalZipFile, UnzipTool.class)
    .toDir(browserStackLocalDirectory.getPath())
    .execute()
`

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
```diff
@@ -0,0 +1,257 @@
...
+
+        try {
+            process = processBuilder.start();
+
+            final Reader reader = new Reader();
+            reader.start();
+            Runtime.getRuntime().addShutdownHook(new CloseChildProcess());
+            countDownLatch.await(20, TimeUnit.SECONDS);
+
+        } catch (UnsupportedEncodingException e) {
```
smiklosovic: are you still on Java 6? What about catching them all at once? Otherwise, just printing stacktrace is not the best thing to do ...

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/PlatformUtils.java
```diff
@@ -0,0 +1,87 @@
+/**
```
smiklosovic: it seems to me you are reinventing a wheel here, this all could be thrown away, use something like this, it is from apache commons lang, it know almost everything out of the box

https://github.com/arquillian/arquillian-cube/commit/195d36daf884d02eabbf3963b35f6deb1c9f210b

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriver.java
```diff
@@ -0,0 +1,65 @@
...
+        this.isSetBrowserStackLocalManaged = isSetBrowserStackLocalManaged;
+    }
+
+    @Override
+    public void get(String url) {
+        String host = null;
+        try {
+            host = new URL(url).getHost();
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
```
smiklosovic: i don't understand what is printed stack trace good for. either log it and take care of it or propagate that exception higher. on line 55 you are checking if host != null which will be in this case so it will fail somewhere else ...

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+public class BrowserStackDriverFactory implements
+    Configurator<BrowserStackDriver, WebDriverConfiguration>,
+    Instantiator<BrowserStackDriver, WebDriverConfiguration>,
+    Destructor<BrowserStackDriver> {
+
+    private static final Logger log = Logger.getLogger(BrowserStackDriverFactory.class.getName());
+
+    @Inject
+    protected Instance<BrowserCapabilitiesRegistry> registryInstance;
```
smiklosovic: are you sure it has to be protected? i would say private is the keyword you want.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
```diff
@@ -0,0 +1,130 @@
...
+        browserStackDriver.quit();
+    }
+
+    public BrowserStackDriver createInstance(WebDriverConfiguration configuration) {
+        try {
+            Capabilities capabilities = configuration.getCapabilities();
+
+            String url = (String) capabilities.getCapability(URL);
+            String accessKey = null;
+            if (isEmpty(url)) {
```
smiklosovic: isNullOrEmpty

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
```diff
@@ -0,0 +1,130 @@
...
+                String username = (String) capabilities.getCapability(USERNAME);
+                accessKey = (String) capabilities.getCapability(ACCESS_KEY);
+                if (isEmpty(accessKey)) {
+                    accessKey = (String) capabilities.getCapability("automate.key");
+                }
+                if (isEmpty(username) || isEmpty(accessKey)) {
+                    log.severe(
```
smiklosovic: I would throw it here in case you can not deal with it anymore, returning null is just weird.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
```diff
@@ -0,0 +1,130 @@
...
+                        "You have to specify either an username and an access.key or the whole url in your arquillian descriptor");
+                    return null;
+                } else {
+                    url = "http://" + username + ":" + accessKey + "@hub.browserstack.com/wd/hub";
```
smiklosovic: you could use String.format()

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
```diff
@@ -0,0 +1,130 @@
...
+                }
+            }
+
+            boolean isSetBrowserStackLocal = capabilities.is(BROWSERSTACK_LOCAL);
+            boolean isSetBrowserStackLocalManaged = capabilities.is(BROWSERSTACK_LOCAL_MANAGED);
+
+            if (isSetBrowserStackLocal && isSetBrowserStackLocalManaged && (!isEmpty(accessKey) || !isEmpty(url))) {
```
smiklosovic: I had to double read it, it is rather hard to follow this logic, I realized that url will not be null because if access key is null, this main if would never be entered in the first place.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriver.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        if (host != null && ("localhost" .equals(host) || "127.0.0.1" .equals(host))) {
```
smiklosovic: unnecessary spaces around .equals methods

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
```diff
@@ -0,0 +1,257 @@
...
+    private boolean isEmpty(String object) {
+        return object == null || object.isEmpty();
+    }
+
+    /**
+     * This thread reads an output from the BrowserStackLocal binary and prints it on the standard output. At the same
+     * time it checks if the output contains one of the strings that indicate that the binary has been successfully
+     * started and the connection established or that another BrowserStackLocal binary is already running
+     */
+    private class Reader extends Thread {
```
smiklosovic: I do not want to go into details here. I do not have code open in ide and if it works to your satisfaction, I would just leave it as it is. I could be _somehow_ rewritten into Spacelift (not sure if it is possible) but I think it is just not necessary. If it works, just keep it here :)

drone-bom/pom.xml
MatousJobanek: Same as for Sauce Labs - this part will be merged, so there will be two extensions

extension/arquillian-drone-browserstack-extension/README.adoc
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/README.adoc
MatousJobanek: BTW. with the arquillian-core 1.1.9.Final and newer it should be possible to specify it on the command line as well: http://arquillian.org/blog/2015/09/02/arquillian-core-1-1-9-Final/

extension/arquillian-drone-browserstack-extension/pom.xml
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
```diff
@@ -0,0 +1,130 @@
...
+    Configurator<BrowserStackDriver, WebDriverConfiguration>,
+    Instantiator<BrowserStackDriver, WebDriverConfiguration>,
+    Destructor<BrowserStackDriver> {
+
+    private static final Logger log = Logger.getLogger(BrowserStackDriverFactory.class.getName());
+
+    @Inject
+    protected Instance<BrowserCapabilitiesRegistry> registryInstance;
+
+    public WebDriverConfiguration createConfiguration(ArquillianDescriptor arquillianDescriptor,
```
smiklosovic: do not these methods override methods in interfaces? what about adding `@Override` annotations on them?

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
```diff
@@ -0,0 +1,130 @@
...
+                    url = "http://" + username + ":" + accessKey + "@hub.browserstack.com/wd/hub";
+                }
+            }
+
+            boolean isSetBrowserStackLocal = capabilities.is(BROWSERSTACK_LOCAL);
+            boolean isSetBrowserStackLocalManaged = capabilities.is(BROWSERSTACK_LOCAL_MANAGED);
+
+            if (isSetBrowserStackLocal && isSetBrowserStackLocalManaged && (!isEmpty(accessKey) || !isEmpty(url))) {
+                if (isEmpty(accessKey)) {
+                    accessKey = url.substring(url.lastIndexOf(":") + 1, url.indexOf("@"));
```
smiklosovic: you could potentially do some matcher for url. you expect that it has some format which does not need to be true every time. if that url would be of wrong format, you would throw ex or log it.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
MatousJobanek: Yes, but it is an exception with a meaningful message, so the user should understand where the problem is.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
MatousJobanek: copy-paste from Drone implementation, but Yes, you are right :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
MatousJobanek: I was trying to rewrite it into Spacelift, but without any success :-(. I've discussed this code with some guys from the office and I think that it should work.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/PlatformUtils.java
MatousJobanek: copy-paste from Drone implementation, but Yes, you are right :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriver.java
MatousJobanek: I thought that the stacktrace gives more information when something fails. But yeah, you are right that in this case same warning log would be more suitable.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriver.java
MatousJobanek: spaces :+1: 
equals: http://stackoverflow.com/questions/9888508/string-equals-argument-ordering
It's true, that in this case the host won't be null - it's just habit to write it in this way.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
MatousJobanek: copy-paste from Drone implementation, but Yes, you are right :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
MatousJobanek: In my view, this seems to be more readable

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
MatousJobanek: :-D haven't realised that. Thanks :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/local/BrowserStackLocalRunner.java
MatousJobanek: I would keep it as it is because of lazy initialization. This class is created only if the user has appropriate property set to true, otherwise it won't be created.

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriver.java
MatousJobanek: so, the spaces were added by IDEA and I haven't found a way to change the formatting... so, I'll change the order in equals statement to get rid of these white spaces :-/

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
MatousJobanek: :+1:

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriverFactory.java
MatousJobanek: I don't want to do some strict checker of url format - I don't know what options are possible in BrowserStack/Sauce Labs. For example if you would run your own instance (if it is possible at all). In this case you could use your own url format and do not use managed binary mode and everything would work without any problem... theoretically :-)

extension/arquillian-drone-browserstack-extension/src/main/java/org/arquillian/drone/browserstack/extension/webdriver/BrowserStackDriver.java
TadeasKriz: I myself don't like the switched order of equals. Especially when you also check for host not being nil.

MatousJobanek: Already fixed: https://github.com/arquillian/arquillian-extension-drone/pull/62/commits/8e428516070745e0274aa7fafeb47aef9cbcb24c

MatousJobanek:
Landed in https://github.com/arquillian/arquillian-extension-drone/commit/137342c8424ec07172e2decbd18e7a98a41c86bf

Issue #64, opened 2016-07-29
Cannot use chromeOptions in the arquillian.xml


opichals:
To configure the drone to provide performance data I need to set chromeOptions capabilities as follows ([complete js webdriver example](http://codegists.com/snippet/javascript/chrome-selenium-timelinejs_axemclion_javascript)):

```
{
    "browserName": "chrome",
    "chromeOptions": {
        "perfLoggingPrefs": {
            "traceCategories": ",blink.console,disabled-by-default-devtools.timeline,benchmark"
        },
        "args": ["--enable-gpu-benchmarking", "--enable-thread-composting"]
    },
    "loggingPrefs": {
        "performance": "ALL"
    }
}
```

It appears that this cannot be easily done in arquillian.xml.

opichals:
https://issues.jboss.org/browse/ARQ-2037

MatousJobanek:
I sent a PR. Please take a look if it works in your case and if it is implemented in a suitable way.
https://github.com/arquillian/arquillian-extension-drone/pull/65
Thanks

MatousJobanek:
Closing this issue as fixed by a1d3e46

michalvanco:
I just tried this change for our use-case and it works perfectly! Thank you @MatousJobanek ;)

kvalev:
It seems that the [documentation](http://arquillian.org/arquillian-extension-drone/#configuring-drone-instances) is out of date, as it does not mention anything about this feature.

MatousJobanek:
@kvalev Thank you for pointing this out - I've added the information about the ChromeOptions only into the release announcement: http://arquillian.org/blog/2016/09/14/arquillian-extension-drone-2-0-1-Final/, but obviously forgot to add to the documentation. I will fix it.

MatousJobanek:
@kvalev fixed http://arquillian.org/arquillian-extension-drone/#chrome-options
thanks again for reporting this

PR #65, opened 2016-09-02
ARQ-2037 Added a support of additional browser capabilities for Chrome

MatousJobanek:
With this commit, it should be possible to set all possible chrome options into
a chromeDriver through arquillian.xml file. All the options you can find
here: https://sites.google.com/a/chromium.org/chromedriver/capabilities
All the options are set into ChromeOptions class:
https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/chrome/ChromeOptions.html
which means that the parameter names tightly depend on names of the set/add
methods.
It is expected that the name of a parameter consist of:
`"chrome" + (name of set/add method of ChromeOption class without first three chars)`
whole string in camelcase.
eg for option args there should be:
`<property name="chromeArguments">--your-cool --arguments</property>`
If the value can be an array or list of strings/files, then specify all of
them in one string separated by space. (This is also applied for
extensions as well as for encoded extensions.)
In the case of experimental options it is a little bit different. As there
is expected a set of dictionaries, the most suitable way is using JSON
format - eg:
`<property name="chromeExperimentalOption">{"perfLoggingPrefs": {
"traceCategories":
",blink.console,disabled-by-default-devtools.timeline,benchmark"
 }, "prefs": {"download.default_directory":
"/usr/local/path/to/download/directory"} }</property>`
(can be in multiline format )

If you are still struglling with passing required options through
arquillian.xml, you can use a parameter chromePrintOptions with a value
`true`:
`<property name="chromePrintOptions">true</property>`
This ensures that Drone prints whole content of ChromeOptions in a JSON
format on a standard output.

This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-extension-drone/65)

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/CapabilitiesMapper.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/CapabilitiesMapper.java
@@ -1,0 +1,192 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat Middleware LLC, and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.arquillian.drone.webdriver.factory;
+
+import java.io.File;
+import java.lang.reflect.Array;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.reflect.TypeToken;
+import org.jboss.arquillian.drone.webdriver.utils.StringUtils;
+import org.openqa.selenium.remote.DesiredCapabilities;
+
+/**
+ * @author <a href="mailto:mjobanek@redhat.com">Matous Jobanek</a>
+ */
+public class CapabilitiesMapper {
+
+    /**
+     * Parses capabilities set in {@link DesiredCapabilities} and according to set-method names it sets the values into
+     * corresponding variables of the given Object instance. It is expected that the parameters defined in arquillian.xml
+     * file have a specific browserPrefix; after this prefix then there is the parameter name itself
+     * (whole string has to be in camelcase)
+     *
+     * @param object An instance of an object the values should be set into
+     * @param capabilities A {@link DesiredCapabilities} that contains parameters and its values set in arquillian.xml
+     * @param browserPrefix A prefix the should the mapped parameters should start with
+     */
+    public static void mapCapabilities(Object object, DesiredCapabilities capabilities, String browserPrefix) {
+
+        Method[] methods = object.getClass().getMethods();
+        List<String> processedMethods = new ArrayList<String>();
+
+        for (Method method : methods) {
+            if (isSetter(method)) {
+
+                method.setAccessible(true);
+                String methodName = method.getName();
+                if (processedMethods.contains(methodName)) {
+                    continue;
+                }
+
+                String propertyName = browserPrefix + methodName.substring(3);
+                propertyName = Character.toLowerCase(propertyName.charAt(0)) + propertyName.substring(1);
+                String capability = (String) capabilities.getCapability(propertyName);
+
+                if (Validate.nonEmpty(capability)) {
+                    try {
+                        if (method.getParameterTypes().length == 1) {
+                            Object converted = convert(method, capability);
+                            if (converted != null) {
+                                method.invoke(object, converted);
+                                processedMethods.add(methodName);
+                            }
+
+                        } else if (shouldContainDictionaries(method)) {
+                            handleDictionaries(object, method, capability);
+                            processedMethods.add(methodName);
+                        }
+
+                    } catch (Exception e) {
+                        e.printStackTrace();
+                    }
+                }
+            }
+        }
+    }
+
+    private static void handleDictionaries(Object object, Method method, String capability)
+        throws InvocationTargetException, IllegalAccessException {
+
+        String trimmedCapability = StringUtils.trimMultiline(capability);
+        JsonObject json = new JsonParser().parse(trimmedCapability).getAsJsonObject();
+        Set<Map.Entry<String, JsonElement>> entries = json.entrySet();
+        final Type type = new TypeToken<Map<String, String>>() {
+        }.getType();
+
+        for (Map.Entry<String, JsonElement> entry : entries) {
+            String key = entry.getKey();
+            Map<String, String> values = new Gson().fromJson(entry.getValue(), type);
+
+            method.invoke(object, key, values);
+        }
+    }
+
+    private static boolean shouldContainDictionaries(Method method) {
+        return
+            (method.getParameterTypes().length == 2)
+                &&
+                (method.getParameterTypes()[0].isAssignableFrom(String.class)
+                    || method.getParameterTypes()[0].isAssignableFrom(Object.class))
+                &&
+                (method.getParameterTypes()[1].isAssignableFrom(Object.class)
+                    || method.getParameterTypes()[1].isAssignableFrom(Map.class));
+    }
+
+    private static Object convert(Method method, String capability) {
+        Class<?> parameterType = method.getParameterTypes()[0];
+        Object converted = null;
+
+        if ((converted = booleanNumberStringOrFile(parameterType, capability)) != null) {
+            return converted;
+
+        } else if (parameterType.isArray()) {
+            return handleArray(parameterType.getComponentType(), capability);
+
+        } else if (parameterType.isAssignableFrom(List.class)) {
+            return handleList(method, capability);
+        }
+
+        return null;
+    }
+
+    private static <T> T[] handleArray(Class<T> parameterType, String capability){
+        List<T> convertedList = getConvertedList(parameterType, capability);
+        T[] array = (T[]) Array.newInstance(parameterType, convertedList.size());
+        return convertedList.toArray(array);
+    }
+
+    private static Object handleList(Method method, String capability) {
+
+        Type[] genericParameterTypes = method.getGenericParameterTypes();
+        if (genericParameterTypes.length == 1) {
+            Type type = genericParameterTypes[0];
+
+            if (type instanceof ParameterizedType) {
+                Type[] parameters = ((ParameterizedType) type).getActualTypeArguments();
+                if (parameters.length == 1) {
+                    return getConvertedList((Class<Object>) parameters[0], capability);
+                }
+            }
+        }
+        return null;
+    }
+
+    private static <T> List<T> getConvertedList(Class<T> parameter, String capability) {
+
+        List<String> values = StringUtils.tokenize(capability);
+        List<T> convertedList = new ArrayList<T>(values.size());
+        for (String value : values) {
+
+            convertedList.add(booleanNumberStringOrFile(parameter, value));
+        }
+
+        return convertedList;
+    }
+
+    private static <T> T booleanNumberStringOrFile(Class<T> clazz, String value) {
+        return (T) (!String.class.equals(clazz) ?
+                    (!File.class.equals(clazz) ? (
+                        !Integer.class.equals(clazz) && !Integer.TYPE.equals(clazz) ? (
+                            !Double.class.equals(clazz) && !Double.TYPE.equals(clazz) ? (
+                                !Long.class.equals(clazz) && !Long.TYPE.equals(clazz) ? (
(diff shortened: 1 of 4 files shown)


MatousJobanek:
This is a fix for #64

TadeasKriz:
Review status: 0 of 7 files reviewed at latest revision, 9 unresolved discussions.

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/CapabilitiesMapper.java, line 173 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp9RPi6i1DsgJSuYJQ:-KQp9RPi6i1DsgJSuYJR:b3hscep) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/CapabilitiesMapper.java#L173)):*
> ```Java
> 
>     private static <T> T booleanNumberStringOrFile(Class<T> clazz, String value) {
>         return (T) (!String.class.equals(clazz) ?
> ```

This method is really cluttered and confusing. Wouldn't it be better to not use nesting/ternary operators? I.e.:
```
if (File.class.equals(clazz)) {
    return new File(value)
}
if (Integer.class.equals(clazz)) {
    return Integer.valueOf(value)
}

// the same for other types

return null
```

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java, line 116 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp6XNRRtYKSBo0AH4s:-KQp6XNRRtYKSBo0AH4t:b7694k) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java#L116)):*
> ```Java
>         if (Validate.nonEmpty(printChromeOptions) && Boolean.valueOf(printChromeOptions.trim())){
>             try {
>                 System.out.println("======== Chrome options =========");
> ```

Shouldn't some logger be used instead of writing directly to out?

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java, line 120 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp95y3JgEJX_fOsWhP:-KQp95y3JgEJX_fOsWhQ:b-sv4n1i) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java#L120)):*
> ```Java
>                 System.out.println("===== End of Chrome options =====");
>             } catch (IOException e) {
>                 e.printStackTrace();
> ```

Who is throwing the `IOException` here? Is it really possible to ignore it and is it wise to print the stack trace?

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 58 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp8MPNaTMxn-pLzvKa:-KQp8MPNaTMxn-pLzvKb:b1obmug) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L58)):*
> ```Java
>     }
> 
>     public static String trimMultiline(String toTrim) {
> ```

This whole method looks like it was taken from some disassembled Java code.

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 60 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp7C979qQ6ZwOe8pvM:-KQp7C98XjKozFL8RPIu:b-sbx46k) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L60)):*
> ```Java
>     public static String trimMultiline(String toTrim) {
>         StringBuilder builder = new StringBuilder(toTrim.length());
>         String[] var2 = toTrim.split("\\s+");
> ```

More descriptive variable names would help in this method.

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 60 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp7pP2Rz1btNkibD7Q:-KQp7pP2Rz1btNkibD7R:b413t7r) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L60)):*
> ```Java
>     public static String trimMultiline(String toTrim) {
>         StringBuilder builder = new StringBuilder(toTrim.length());
>         String[] var2 = toTrim.split("\\s+");
> ```

Correct me if I'm wrong, but doesn't this split by every whitespace, removing the need to do `.trim()`? Also I though this method is trimming spaces on every line. So if it trims each space-separated part, I suggest renaming the method.

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 63 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp7Sy_yDJObM_CXimL:-KQp7Sy_yDJObM_CXimM:b-c38ta6) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L63)):*
> ```Java
>         int var3 = var2.length;
> 
>         for(int var4 = 0; var4 < var3; ++var4) {
> ```

Why not iterate using `for (String token : var2) { .. }`?

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 65 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp7zHgvzKzSesioGT4:-KQp7zHgvzKzSesioGT5:b-lynmlf) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L65)):*
> ```Java
>         for(int var4 = 0; var4 < var3; ++var4) {
>             String token = var2[var4];
>             if(token != null && !"".equals(token.trim())) {
> ```

I'm not sure about the performance of `.trim` method, but it might be better to extract it to a variable.

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 70 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp8Gm50NdKf3J12tt6:-KQp8Gm50NdKf3J12tt7:b-b0ugmf) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L70)):*
> ```Java
>         }
> 
>         return builder.toString().trim();
> ```

Isn't this trim unnecessary?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65)*

MatousJobanek:
Review status: 0 of 7 files reviewed at latest revision, 9 unresolved discussions.

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java, line 120 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp95y3JgEJX_fOsWhP:-KQsks5av-EXISS7tzNV:b-vocy6s) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java#L120)):*
<details><summary><i>Previously, TadeasKriz (Tadeas Kriz) wrote…</i></summary>
> Who is throwing the `IOException` here? Is it really possible to ignore it and is it wise to print the stack trace?
</details>
The IOException is thrown by chromeOptions.toJson(). So I think that it is possible to ignore it and regarding printing stacktrace - I put it there to let user know when something bad happened

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 58 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp8MPNaTMxn-pLzvKa:-KQslgpPVnXaZVvPzxHb:b-azyfj3) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L58)):*
<details><summary><i>Previously, TadeasKriz (Tadeas Kriz) wrote…</i></summary>
> This whole method looks like it was taken from some disassembled Java code.
</details>
Yeah, I've stolen it from the arquillian-core project: https://github.com/arquillian/arquillian-core/blob/master/container/impl-base/src/main/java/org/jboss/arquillian/container/impl/MultilineTrimmer.java
I've copy-pasted it from decompiled java code in IDE (just wanted to try it if and how it works), so this is the reason why the whole method looks so weird. And after implementing the main code itself I completely forgot to replace it by normal code. Yeah, I'm stupid... I'll fix it

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 60 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp7pP2Rz1btNkibD7Q:-KQsmu9v8Gku0EtdGG9u:bxivrsc) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L60)):*
<details><summary><i>Previously, TadeasKriz (Tadeas Kriz) wrote…</i></summary>
> Correct me if I'm wrong, but doesn't this split by every whitespace, removing the need to do `.trim()`? Also I though this method is trimming spaces on every line. So if it trims each space-separated part, I suggest renaming the method.
</details>
I think that from the original code: https://github.com/arquillian/arquillian-core/blob/master/container/impl-base/src/main/java/org/jboss/arquillian/container/impl/MultilineTrimmer.java it is more clear what it really does. 

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 63 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp7Sy_yDJObM_CXimL:-KQsoFzdxN7cw_IU2awu:b-h8wnh2) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L63)):*
<details><summary><i>Previously, TadeasKriz (Tadeas Kriz) wrote…</i></summary>
> Why not iterate using `for (String token : var2) { .. }`?
</details>
Sorry for this method - as I wrote above - it is my big fault that I forgot to replace it by normal code. 

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 65 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp7zHgvzKzSesioGT4:-KQsnkasmqUivnSFJl7Z:bbv013t) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L65)):*
<details><summary><i>Previously, TadeasKriz (Tadeas Kriz) wrote…</i></summary>
> I'm not sure about the performance of `.trim` method, but it might be better to extract it to a variable.
</details>
Do you have any statistic/reference/article about it?

---

*[drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java, line 70 \[r1\]](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65#-KQp8Gm50NdKf3J12tt6:-KQsnSDiNN9bTJxVybj7:b-qu18v9) ([raw file](https://github.com/arquillian/arquillian-extension-drone/blob/eea6f3778b9fa5e38b810ac615cc09c0d77ebb2e/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/StringUtils.java#L70)):*
<details><summary><i>Previously, TadeasKriz (Tadeas Kriz) wrote…</i></summary>
> Isn't this trim unnecessary?
</details>
Yes it is - the trim is for the last string that is appended with a white space after it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/arquillian/arquillian-extension-drone/65)*

MatousJobanek:
@TadeasKriz thanks a lot for your review and sorry for my stupidity that I forgot to replace the confusing copy-pasted methods.

MatousJobanek:
Landed in https://github.com/arquillian/arquillian-extension-drone/commit/a1d3e46b7134dd0f6cf2425f1d204977416b7651

michalvanco:
Thanks @MatousJobanek for implementing this support!

PR #67, opened 2016-09-05
ARQ-2044 Exposed logic that creates browser capabilities

MatousJobanek:
These methods can be reused for example in Cube integration with Drone. WDYT @lordofthejars

---
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-extension-drone/67)

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
@@ -78,6 +78,20 @@ public class ChromeDriverFactory extends AbstractWebDriverFactory<ChromeDriver> 
     @Override
     public ChromeDriver createInstance(WebDriverConfiguration configuration) {
 
+        Capabilities capabilities = getCapabilities(configuration);
+
+        return SecurityActions.newInstance(configuration.getImplementationClass(), new Class<?>[] { Capabilities.class },
+                new Object[] { capabilities }, ChromeDriver.class);
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


lordofthejars:
The `getCapabilities` right?

MatousJobanek:
Yes, exactly.

lordofthejars:
Yes, partially since in Cube it uses the Remote instance. Also when you say `It also validates if the defined chrome.binary is an executable binary` in case of Cube this will not be installed but it is not erroneous.

bartoszmajsak:
LGTM, except that I don't see any unit tests :)

MatousJobanek:
Well, yeah but it validates only if it is set. But you are right - this would lead to: 
```
       try {
            Capabilities capabilities = getCapabilities(configuration);
        }catch (Exception e){
        }
```
which is not good. Any idea?

lordofthejars:
Maybe adding a boolean or creating another method where paths are checked or not depending on boolean/method.

MatousJobanek:
Yeah, I had been thinking about adding a boolean parameter, but I wanted to avoid it because I wasn't able to decide what to do when the validations are not performed - should the value be set to capabilities anyway or not...? In addition, I wanted to keep the method same across all factory class (kind of API method) but in most classes the parameter is not necessary...
So I pushed my proposition: https://github.com/arquillian/arquillian-extension-drone/pull/67/commits/aac10b5372cad13423a2d56d3f14bce2fb332f74
WDYT?

MatousJobanek:
Landed in https://github.com/arquillian/arquillian-extension-drone/commit/738c45b132ae1a05ee3d6d0002c482665919ca94

PR #68, opened 2016-12-07
Drone & Selenium 3

MatousJobanek:
PR containing changes for new Drone 3
- the main change is an upgrade to Selenim 3
    - after this upgrade the lowest required version of java is 8, so this is reason for new major release of drone
    - there has been also removed module drone-selenium-server which provided a support for deprecated selenium RC - this one has been removed in Selenium 3


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/remote/reusable/ReusableRemoteWebDriver.java
```diff
@@ -80,8 +81,20 @@ public static RemoteWebDriver fromReusedSession(URL remoteAddress, Capabilities
...
+    }
+
+    protected ReusableRemoteWebDriver(URL remoteAddress, Capabilities capabilities, SessionId sessionId) {
+        super();
+        HttpCommandExecutor httpCommandExecutor = new HttpCommandExecutor(remoteAddress);
+        setCommandExecutor(httpCommandExecutor);
         setReusedCapabilities(capabilities);
+
+        setValueToFieldInHttpCommandExecutor(httpCommandExecutor, "commandCodec", Dialect.OSS.getCommandCodec());
+        setValueToFieldInHttpCommandExecutor(httpCommandExecutor, "responseCodec", Dialect.OSS.getResponseCodec());
```
bartoszmajsak: Why do we have to fallback to reflection for this settings?

pom.xml
```diff
@@ -37,22 +37,21 @@
     <properties>
         <!-- Versioning -->
         <version.arquillian.core>1.1.11.Final</version.arquillian.core>
-        <version.com.opera>1.5</version.com.opera>
         <version.phantom.driver>1.2.1.Final</version.phantom.driver>
-        <version.selenium>2.53.1</version.selenium>
-        <version.htmlunit.driver>2.20</version.htmlunit.driver>
+        <version.selenium>3.0.1</version.selenium>
+        <version.htmlunit.driver>2.23</version.htmlunit.driver>
         <version.junit>4.11</version.junit>
```
bartoszmajsak: Can we bump it to `4.12`?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/remote/reusable/ReusableRemoteWebDriver.java
MatousJobanek: Because of this commit: https://github.com/SeleniumHQ/selenium/commit/d725150ac7a9ca5c250b075d66cf7a3caeb32413
The executors are required not to be null when some command (other than the one for creating a new session) is executed: https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/HttpCommandExecutor.java#L152
but these executors are created only when a new session is being created - which is not the case when we are using "reused session". I tried to create a new session and then to replace the sessionId with the older one, but then there was hanging the additional sessions which caused failures.

pom.xml
MatousJobanek: well, yes we can but... in the gradle configuration we are testing against all combination of arquillian-core since 1.1.4.Final; and in the versions 1.1.4 and 1.1.5, there is an issue (when you use junit 4.12) related to an annotation @InSequence. 
So we can, but we need to limit the combination of arquillian-core versions we are testing with. There are two questions now:
1) do we need to test it with these two old versions of arq-core?
2) what would be the improvement when we used junit 4.12 ?

bartoszmajsak: Well, latest core is on this version - hence my question

bartoszmajsak: Yeah, I've noticed these two things. But then still, shouldn't we be relying on what particular `arq-core` ships? I guess that's what we do for `gradle` multi build. So this change in `pom.xml` won't affect it? Or does it take precedence? Imagine `junit:4.13` introduces some breaking changes (we already rely on deprecated methods in `arq-core` so it's more the question of when than if) - this build will have to take this into consideratiojn.

MatousJobanek: Any breaking changes are definitely more important then testing with old arq-core versions.

> this build will have to take this into consideratiojn

you are right about this :-)

Ok, I'll append a commit with related changes ;-)

bartoszmajsak:
@MatousJobanek feel free to squash and merge :)

MatousJobanek:
After a discussion on internal channel, we decided not to go for a new major release, but include these changes in 2.1.0 release.

MatousJobanek:
Landed in https://github.com/arquillian/arquillian-extension-drone/commit/250f206b1a0529190baa1264baad736c8f0c5aee and in following commits

PR #70, opened 2016-12-23
ARQ-1695 Integrated Spacelift into Drone: downloading, extracting and running processes

MatousJobanek:
There where introduced a new abstractions:
ExternalBinary = represents a binary from an external source (eg.: web-driver or selenium server jar)
ExternalBinarySource = represents a an external source a binary could be retrieved from (eg: google storage for selenium bits and chrome webDriver)
BinaryHandler = Class that handles all necessary stuff around binaries (setting and getting properties, downloading, extracting, running, etc...)

So far, the supported binaries (from automatic downloading point of view):
- gecko driver
- chrome driver
- selenium server
- internet explorer driver


vinsguru:
Very nice idea! Great :+1:

MatousJobanek:
Landed in https://github.com/arquillian/arquillian-extension-drone/commit/fb36d41afe9c6de289e5a709a6917c5f7e2ba8a3

Issue #71, opened 2017-01-30
Possibility to add command line arguments for Selenium Server that is run internally.


hemanik:
##### Issue Overview

Currently, if there is no selenium server running on localhost then Drone downloads and runs an instance of selenium server on its own. 

This instance of selenium server is run as a standalone with only [one parameter: -port] that is taken from defined remoteAddress. 

The user should have the possibility to define some additional arguments [ example: -debug -role hub -browserTimeout 1000] for the selenium server.

PR #72, opened 2017-01-30
#71: Adds command line arguments for selenium server.

hemanik:

Adds the possibility to allow user to specify additional arguments [example: -debug -role hub -browserTimeout 1000] to the selenium server that is run internally

- Adds `seleniumServerArgs` as a new `arquillian.xml` property which would be a simple string with arguments  defined.
- Appends the defined arguments to the build command that launches the selenium server.
- Updates necessary tests to include the new property.

**Fixes**: #71

### DIFF
--- a/drone-configuration/src/test/java/org/jboss/arquillian/drone/configuration/MockDroneConfiguration.java
+++ b/drone-configuration/src/test/java/org/jboss/arquillian/drone/configuration/MockDroneConfiguration.java
@@ -155,8 +155,12 @@ public class MockDroneConfiguration implements DroneConfiguration<MockDroneConfi
         this.browserCapabilities = browserCapabilities;
     }
 
-    public String getSeleniumServerArgs() { return seleniumServerArgs; }
+    public String getSeleniumServerArgs() {
+        return seleniumServerArgs;
+    }
 
-    public void setSeleniumServerArgs(String seleniumServerArgs) { this.seleniumServerArgs = seleniumServerArgs; }
+    public void setSeleniumServerArgs(String seleniumServerArgs) {
+        this.seleniumServerArgs = seleniumServerArgs;
+    }
 
 }
--- a/drone-configuration/src/test/java/org/jboss/arquillian/drone/configuration/MockDroneConfiguration.java
+++ b/drone-configuration/src/test/java/org/jboss/arquillian/drone/configuration/MockDroneConfiguration.java
@@ -52,6 +52,8 @@ public class MockDroneConfiguration implements DroneConfiguration<MockDroneConfi
 
     private String browserCapabilities;
 
+    private String seleniumServerArgs;
+
     /*
      * (non-Javadoc)
      *
@@ -152,5 +154,9 @@ public class MockDroneConfiguration implements DroneConfiguration<MockDroneConfi
     public void setBrowserCapabilities(String browserCapabilities) {
         this.browserCapabilities = browserCapabilities;
     }
+
+    public String getSeleniumServerArgs() { return seleniumServerArgs; }
+
+    public void setSeleniumServerArgs(String seleniumServerArgs) { this.seleniumServerArgs = seleniumServerArgs; }
 
 }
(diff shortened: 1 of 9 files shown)


bartoszmajsak:
Is there any way to test those changes automatically?

MatousJobanek:
Good job @hemanik the code looks good!
It would be great to add some information into the README file so users know how to use it. There is only short mention about running selenium server in the paragraph related to tests: https://github.com/arquillian/arquillian-extension-drone#manual-way-to-run-test-suite 
Feel free to create a new paragraph and just shortly describe how to use this feature.
Thanks

hemanik:
@bartoszmajsak, I tried writing tests for the proposed changes but it seems it is not possible to write real unit tests for it. The [related lines of code](https://github.com/arquillian/arquillian-extension-drone/blob/master/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/process/SeleniumServerExecutor.java#L44) make it difficult to test this code since everything is handled within this section.

Is there a better way that you can suggest to test this ?

bartoszmajsak:
@hemanik I haven't look at the codebase lately but maybe @MatousJobanek can shed some light?

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/process/SeleniumServerTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+
+        final String browser = "chrome";
+        final String seleniumServerArgs = "-debug";
+
+        fire(new StartSeleniumServer(seleniumServerBinary, browser, capabilities, url, seleniumServerArgs));
+
+        final String capturedLog = getTestCapturedLog();
+        final String output = parseLogger(capturedLog);
+
+        Assertions.assertThat(output.contains(seleniumServerArgs));
```
MatousJobanek: I think that the correct format of the assertion should be: 
`Assertions.assertThat(output).contains(seleniumServerArgs);`
otherwise it doesn't verify anything. 
This applies also for all following occurrences in next tests...

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/process/SeleniumServerTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @After
+    public void stopSeleniumServer(){
+        fire(new AfterSuite());
+    }
+
+    private String parseLogger(String capturedLog) {
+
+        final String expectedLogPart = "Running Selenium server process: java .*$";
```
MatousJobanek: Technically, your tests only verify that the command was constructed and printed on the stdout. Apart from that, I would also expect a verification that the selenium server reflects arguments that are used - the command was correctly used. 
For this purpose, the easiest way is (FMPOV) using `-role` argument with either `node` or `hub` value. The log printed by the selenium server on the output contains appropriate information.

drone-configuration/src/test/java/org/jboss/arquillian/drone/configuration/MockDroneConfiguration.java
```diff
@@ -153,4 +155,8 @@ public void setBrowserCapabilities(String browserCapabilities) {
         this.browserCapabilities = browserCapabilities;
     }
 
+    public String getSeleniumServerArgs() { return seleniumServerArgs; }
```
MatousJobanek: please, don't use this formatting - keep the formatting consistent.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/process/SeleniumServerTestCase.java
```diff
@@ -0,0 +1,136 @@
...
+        customLogHandler.flush();
+        return logCapturingStream.toString();
+    }
+
+    private void setUpStreams() {
+        System.setOut(new PrintStream(outContent));
+    }
+
+    private void cleanUpStreams() {
+        System.setOut(null);
```
MatousJobanek: I think that this is not a clear way to reset values to default - I think, that there should be something like: 
`System.setOut(System.out);`

MatousJobanek:
Landed in: https://github.com/arquillian/arquillian-extension-drone/commit/b2be16f40f44fbc09df0bac53b33401c7caab772 🚀 Good job @hemanik

PR #73, opened 2017-02-13
Updates documentation for Safaridriver

dipak-pawar:
Adds documentation why are we not supporting following things for Safari:
* Automatic downloading support
* Download webdriver from specific property


MatousJobanek:
Landed in: https://github.com/arquillian/arquillian-extension-drone/commit/df1a876cc0f05c7e16ca72a50339de42bc6b7a5b 🚀  Thank you for your effort @dipak-pawar 👍

PR #74, opened 2017-02-17
ARQ-2080 remote webdriver is closed even when the session is persisted

MatousJobanek:
otherwise, when the test ReusableRemoteWebDriverTestCase was run, then the browser wouldn't be closed


bartoszmajsak:
Any idea why the build failed?

MatousJobanek:
yeah, I know why it failed - with the change I broke the reusable-session functionality
I'll send another PR in a min with a new implementation

PR #75, opened 2017-02-17
ARQ-2080 Drone keeps the latest RemoteWebDriver whose reusable session was persisted and in AfterClass it tries to destroy it

MatousJobanek:
Implemented a registry ReusableRemoteWebDriverToDestroy where I store the latest driver that was used with a reusable session. Then in AfterClass phase, this webdriver is destroyed. 
The reason is that when the reusable session was persisted, then the remote webdriver instance wasn't destroyed. As a result of this, a browser wasn't closed in tests:
ReusableRemoteWebDriverTestCase
TestReusingRemoteWebDriverSession

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/remote/reusable/ReusableRemoteWebDriverToDestroy.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/remote/reusable/ReusableRemoteWebDriverToDestroy.java
@@ -1,0 +1,38 @@
+package org.jboss.arquillian.drone.webdriver.factory.remote.reusable;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.openqa.selenium.WebDriverException;
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+/**
+ * @author <a href="mailto:mjobanek@redhat.com">Matous Jobanek</a>
+ */
+public class ReusableRemoteWebDriverToDestroy {
+
+    private Logger log = Logger.getLogger(ReusableRemoteWebDriverToDestroy.class.getName());
+
+    private boolean destroyed = false;
+
+    private RemoteWebDriver remoteWebDriver;
+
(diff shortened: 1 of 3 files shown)


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/remote/reusable/ReusableRemoteWebDriverToDestroy.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.openqa.selenium.WebDriverException;
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+/**
+ * @author <a href="mailto:mjobanek@redhat.com">Matous Jobanek</a>
+ */
+public class ReusableRemoteWebDriverToDestroy {
```
bartoszmajsak: Why do we need it if there is already `ReusableRemoteWebDriver`? Can't we solve it just by using it?

MatousJobanek: Good point! I was thinking about the same. I decided to create a new class, because I wanted to use it only for this purpose and to limit confusions. I'm not sure what should be the content (if  I used this general class) in other cases https://github.com/arquillian/arquillian-extension-drone/blob/master/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/RemoteWebDriverFactory.java#L168-L189
If it should be removed when it is destroyed or add additional marker that it is destroyed.
There could be an addition question - why do we keep some kind of registry for `ReusableRemoteWebDriver`s and not for other drivers.

bartoszmajsak: I have no idea :) But their names are very similar... So it's already a bit confusing

PR #82, opened 2017-03-28
PhantomJS is downloaded using Spacelift

MatousJobanek:
Drone doesn't rely on arquillian-phantom-* projects any more
PhantomJS is downloaded using Spacelift (the logic implemented in Drone and used for other binaries) from Bitbucket download storage
The latest version is retrieved from GH tags.

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/AbstractBinaryHandler.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/AbstractBinaryHandler.java
@@ -1,9 +1,5 @@
 package org.jboss.arquillian.drone.webdriver.binary.handler;
 
-import java.io.File;
-import java.net.URL;
-import java.util.logging.Logger;
-
 import org.jboss.arquillian.drone.webdriver.binary.BinaryFilesUtils;
 import org.jboss.arquillian.drone.webdriver.binary.downloading.Downloader;
 import org.jboss.arquillian.drone.webdriver.binary.downloading.ExternalBinary;
@@ -12,6 +8,10 @@ import org.jboss.arquillian.drone.webdriver.utils.Constants;
 import org.jboss.arquillian.drone.webdriver.utils.PropertySecurityAction;
 import org.jboss.arquillian.drone.webdriver.utils.Validate;
 import org.openqa.selenium.remote.DesiredCapabilities;
+
+import java.io.File;
+import java.net.URL;
+import java.util.logging.Logger;
 
 /**
  * Class that handles system properties, properties stored in capabilities, downloading, extracting and setting binaries
@@ -185,7 +185,7 @@ public abstract class AbstractBinaryHandler implements BinaryHandler {
         File downloaded = Downloader.download(targetDir, from);
         File extraction = BinaryFilesUtils.extract(downloaded);
         File[] files = extraction.listFiles(file -> file.isFile());
-        if (files.length == 0) {
+        if (files == null || files.length == 0) {
             throw new IllegalStateException(
                 "The number of extracted files in the directory " + extraction + " is 0. There is no file to use");
         }
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/PhantomJSDriverFactory.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/PhantomJSDriverFactory.java
@@ -16,14 +16,12 @@
  */
 package org.jboss.arquillian.drone.webdriver.factory;
 
-import java.io.IOException;
-
 import org.jboss.arquillian.drone.spi.Configurator;
 import org.jboss.arquillian.drone.spi.Destructor;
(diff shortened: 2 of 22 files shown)


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/AbstractBinaryHandler.java
```diff
@@ -185,7 +185,7 @@ protected File downloadAndPrepare(File targetDir, URL from) throws Exception {
         File downloaded = Downloader.download(targetDir, from);
         File extraction = BinaryFilesUtils.extract(downloaded);
         File[] files = extraction.listFiles(file -> file.isFile());
```
bartoszmajsak: So we are officially JDK8 in drone?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/PhantomJSDriverFactory.java
```diff
@@ -35,17 +33,14 @@
  * PhantomJSDriver.
  *
  * @author <a href="kpiwko@redhat.com">Karel Piwko</a>
- *
+ * @author <a href="mjobanek@redhat.com">Matous Jobanek</a>
```
bartoszmajsak: I was thinking about removing author tags from every class :) WDYT? (nothing for this PR - just to open the discussion)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/Validate.java
```diff
@@ -87,6 +98,71 @@ public static void isExecutable(String path, String message) throws IllegalArgum
...
+        }
+    }
+
+    private static void addMacSpecificPath(final ImmutableSet.Builder<String> pathSegmentBuilder) {
+        File pathFile = new File("/etc/paths");
+        if (pathFile.exists()) {
+            try {
+                pathSegmentBuilder.addAll(Files.readLines(pathFile, Charsets.UTF_8));
+            } catch (IOException e) {
+                // Guess we won't include those, then
```
bartoszmajsak: At least some logging? :)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/AbstractBinaryHandler.java
MatousJobanek: Yes. Selenium 3.x.z is only JDK8 compatible. 
I decided to do it when I released the first alpha version.

bartoszmajsak: We should probably highlight this information somewhere - maybe on the module page of arq.org?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/PhantomJSDriverFactory.java
MatousJobanek: No problem with it. Yeah, the question is if it brings any value; as all the changes is trackable via git, then I don't see benefit.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/Validate.java
MatousJobanek: Yeah, I missed it. Thx

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/PhantomJSDriverFactory.java
bartoszmajsak: Also it gets outdated pretty soon and... hints to code ownership (by individual)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/Validate.java
MatousJobanek: fixed

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/AbstractBinaryHandler.java
MatousJobanek: I don't think that it is necessary. From my point of view, it is something that we should have done a long time ago, I was only waiting till Selenium changes its support of java version first. 
For those who would be really interested in it, it is available in the pom.xml file :-)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/PhantomJSDriverFactory.java
MatousJobanek: Maybe it would be best to remove them as a content of one separate commit (including formatting changes). Jira issue would be great.

PR #84, opened 2017-04-10
Fixing wrong profiles from pom.xml

dipak-pawar:
Fixing pom profiles for htmlUnit

- We have used htmlUnit as a browser in implementation, where at the same time we are using `htmlunit` in profiles

### DIFF
--- a/docs/configuring-drone-instances.adoc
+++ b/docs/configuring-drone-instances.adoc
@@ -75,8 +75,8 @@ Following values are valid:
 
 . chrome +
 . firefox +
-. htmlUnit +
-. internetExplorer +
+. htmlUnit  or htmlunit +
+. internetExplorer or internetexplorer +
 . opera +
 . phantomjs +
 . safari +
--- a/drone-webdriver/pom.xml
+++ b/drone-webdriver/pom.xml
@@ -305,9 +305,12 @@
                     <requireProperty>
                       <property>browser</property>
                       <message>browser property must be specified!</message>
-                      <regex>firefox|android|htmlUnit|safari|chrome|opera|phantomjs|iphone|internetExplorer|edge</regex>
-                      <regexMessage>-Dbrowser must be one of: firefox, android, htmlUnit, safari, chrome, opera,
-                        phantomjs, iphone, internetExplorer, edge but was ${browser}
+                      <regex>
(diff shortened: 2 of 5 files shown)


bartoszmajsak:
I'm not sure what it fixes. We cannot specify `htmlunit` as a browser name for drone?

dipak-pawar:
@bartoszmajsak we have to use htmlUnit not `htmlunit` in drone. Or we have to check for both & make it downcase in implementation

bartoszmajsak:
I think we should support both then and leave profile as it was before

dipak-pawar:
:+1:

bartoszmajsak:
Thanks for spotting it!

bartoszmajsak:
If you are willing to fix it we can use this PR to implement this feature.

dipak-pawar:
cc @MatousJobanek

MatousJobanek:
Thx for the fix - yeah, this is a better solution than just fixing the wrong profile configuration (give us more possibilities). 
However, we will need additional changes in the test suite and in the implementation. 

I would start with the pom with the profiles - you need to cover also the second option: `htmlUnit `

Then, if I run the test suite with `htmlUnit` property, I get several times an Exception:
~~~
java.lang.IllegalStateException: Unable to initialize WebDriver instance. Please specify a valid browser instead of htmlUnit. Available options are: internetExplorer, edge, opera, chrome, safari, phantomjs, firefox, htmlUnit
~~~
This is probably because you didn't change the values in `BrowserCapabilitiesList` and the `WebDriverConfiguration` still returns original string when `getBrowser()` method is called.

If I run with `htmlunit` I get an error for `TestRemoteWebDriverFactorySessionStoring` so you need to modify it accordingly. 

You also get an error for the test case: `TestReusableRemoteWebDriver` - this one seems to be caused by a bug in HtmlUnit so we cannot anyhow influence it. A possible workaround is seting `setThrowExceptionOnScriptError` option of the WebClient to `false` but we need to have the functionality first. For now you can skip the test.

dipak-pawar:
@MatousJobanek Please have a look into this.

MatousJobanek:
Looks good now, thank you @dipak-pawar 👍

PR #85, opened 2017-04-12
Adds webClient options to HtmlUnitDriver - ARQ-1836

dipak-pawar:
You can add webClientOptions to HtmlUnitDriver.

- Subclassed HtmlUnitDriver to override `protected modifyWebClient()`
- Used desired capabilities to set options from confiiguration to `WebClientOption` 

**Fixes**: #ARQ-1836

### DIFF
--- a/docs/configuring-drone-instances.adoc
+++ b/docs/configuring-drone-instances.adoc
@@ -117,6 +117,12 @@ FirefoxDriver
 |Path to Firefox user preferences. This file will be parsed and values
 will be applied to freshly created Firefox profile.
 
+|htmlUnitWebClientOptions
+|-
+|Comma separated list of webClientOptions for HtmlUnitDriver. e.g.
+timeout=10,javaScriptEnabled=true, throwExceptionOnScriptError=false
+
+
 |dimensions
 |-
 |Dimensions of browser window in `widthxheight` format. This will resize
--- a/docs/configuring-drone-instances.adoc
+++ b/docs/configuring-drone-instances.adoc
@@ -119,8 +119,8 @@ will be applied to freshly created Firefox profile.
 
 |htmlUnitWebClientOptions
 |-
-|Comma separated list of webClientOptions for HtmlUnitDriver. e.g.
-timeout=10,javaScriptEnabled=true, throwExceptionOnScriptError=false
+|Semicolon separated list of webClientOptions for HtmlUnitDriver. e.g.
+timeout=10; javaScriptEnabled=true; throwExceptionOnScriptError=false; SSLClientProtocols=protocal1, protocal2
 
 
 |dimensions
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverFactory.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverFactory.java
@@ -16,12 +16,18 @@
  */
 package org.jboss.arquillian.drone.webdriver.factory;
 
+import com.gargoylesoftware.htmlunit.WebClient;
+import com.gargoylesoftware.htmlunit.WebClientOptions;
+import java.util.LinkedHashMap;
+import java.util.Map;
 import org.jboss.arquillian.drone.spi.Configurator;
 import org.jboss.arquillian.drone.spi.Destructor;
 import org.jboss.arquillian.drone.spi.Instantiator;
 import org.jboss.arquillian.drone.webdriver.configuration.WebDriverConfiguration;
+import org.jboss.arquillian.drone.webdriver.htmlunit.HtmlUnitDriver;
+import org.jboss.arquillian.drone.webdriver.utils.Validate;
 import org.openqa.selenium.Capabilities;
-import org.openqa.selenium.htmlunit.HtmlUnitDriver;
+import org.openqa.selenium.remote.DesiredCapabilities;
 
 /**
  * Factory which combines {@link org.jboss.arquillian.drone.spi.Configurator},
@@ -62,8 +68,17 @@ public class HtmlUnitDriverFactory extends AbstractWebDriverFactory<HtmlUnitDriv
     @Override
     public HtmlUnitDriver createInstance(WebDriverConfiguration configuration) {
         Capabilities capabilities = getCapabilities(configuration, true);
-        return SecurityActions.newInstance(configuration.getImplementationClass(), new Class<?>[] {Capabilities.class},
-            new Object[] {capabilities}, HtmlUnitDriver.class);
+        final HtmlUnitDriver htmlUnitDriver =
+            SecurityActions.newInstance(configuration.getImplementationClass(), new Class<?>[] {Capabilities.class},
+                new Object[] {capabilities}, HtmlUnitDriver.class);
+
+        final String htmlUnitClientOptions = (String) capabilities.getCapability("htmlUnitWebClientOptions");
+        if (Validate.nonEmpty(htmlUnitClientOptions)) {
+            WebClient webClient = htmlUnitDriver.getWebClient();
+            setClientOptions(webClient, htmlUnitClientOptions);
+        }
+
+        return htmlUnitDriver;
     }
 
     /**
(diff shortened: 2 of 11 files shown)


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/htmlunit/HtmlUnitDriver.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public HtmlUnitDriver(Capabilities capabilities) {
+        super(capabilities);
+    }
+
+    @Override
+    public WebClient modifyWebClient(WebClient client) {
+        final WebClientOptions webClientOptions = client.getOptions();
+
+        return super.modifyWebClient(client);
```
bartoszmajsak: How should it work? It seems to me that superclass does [nothing](https://github.com/SeleniumHQ/htmlunit-driver/blob/15e9ecb98db58e6913cc781fa04ddb0903407626/src/main/java/org/openqa/selenium/htmlunit/HtmlUnitDriver.java#L488-L491)

dipak-pawar: Currently if we use HtmlDriver from selenium from which we couldn't get webClient. So we have to subclass it to get webclient & set it's options.  

About this method it's not doing anything because we already setup webClientOption [here](https://github.com/arquillian/arquillian-extension-drone/pull/85/files#diff-0771f250fc3a41a9424b87d1373b72a2R106). In future we may need to set some options e.g. CredentialsProvider which we can set when required.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/htmlunit/HtmlUnitDriver.java
```diff
@@ -0,0 +1,23 @@
+package org.jboss.arquillian.drone.webdriver.htmlunit;
+
+import com.gargoylesoftware.htmlunit.WebClient;
+import org.openqa.selenium.Capabilities;
+
+public class HtmlUnitDriver extends org.openqa.selenium.htmlunit.HtmlUnitDriver {
```
MatousJobanek: change the name to not confuse. eg: DroneHtmlUnitDriver

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverFactory.java
```diff
@@ -86,4 +101,22 @@ public Capabilities getCapabilities(WebDriverConfiguration configuration, boolea
...
+
+    private void setClientOptions(WebClient webClient, String htmlUnitWebClientOptions) {
+        final WebClientOptions webClientOptions = webClient.getOptions();
+
+        Map<String, String> clientOptions = new LinkedHashMap<>();
+        final String[] options = htmlUnitWebClientOptions.split(",");
+        for (String option : options) {
+            final String[] keyValue = option.split("=");
+            if (keyValue.length == 2) {
+                String key = keyValue[0].trim();
```
MatousJobanek: Remember that the whole property can be a multiline string - use `StringUtils.trimMultiline` before you split it for the first time

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverFactory.java
```diff
@@ -86,4 +101,22 @@ public Capabilities getCapabilities(WebDriverConfiguration configuration, boolea
...
+                key = BROWSER_CAPABILITIES + Character.toUpperCase(key.charAt(0)) + key.substring(1);
```
MatousJobanek: you could also check the length of the key - not to get an exception when the key is empty

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverFactory.java
```diff
@@ -86,4 +101,22 @@ public Capabilities getCapabilities(WebDriverConfiguration configuration, boolea
     protected String getDriverReadableName() {
         return BROWSER_CAPABILITIES;
     }
+
+    private void setClientOptions(WebClient webClient, String htmlUnitWebClientOptions) {
+        final WebClientOptions webClientOptions = webClient.getOptions();
+
+        Map<String, String> clientOptions = new LinkedHashMap<>();
+        final String[] options = htmlUnitWebClientOptions.split(",");
```
MatousJobanek: what if we want to set an array of strings as a value - eg: http://htmlunit.sourceforge.net/apidocs/com/gargoylesoftware/htmlunit/WebClientOptions.html#setSSLClientProtocols-java.lang.String:A-
correct me if my question is nonsense...

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/htmlunit/HtmlUnitDriver.java
dipak-pawar: :+1:

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverFactory.java
dipak-pawar: :+1:

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverFactory.java
dipak-pawar: I haven't think about this in this implementation. But I believe we need some delimiter for it.

e.
```
<property name=hmtlUnitWebClientOptions>timeout=10, SSLClientProtocols=foo:bar:test</property> 
```
wdyt?

MatousJobanek: how about using semicolon to delimit key-values and coma for args:
~~~xml
<property name=hmtlUnitWebClientOptions>timeout=10;SSLClientProtocols=foo,bar,test</property> 
~~~
from my point of view, it looks more natural

dipak-pawar: :+1:

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverFactory.java
dipak-pawar: :+1:

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/factory/HtmlUnitDriverTest.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class HtmlUnitDriverTest {
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void shouldOverrideWebClientOptions() throws IOException {
```
MatousJobanek: Interesting test.
What I had in my mind was just creating an instance of the DroneHtmlUnitDriver (as you did) and then get the options and verify the content:
~~~java
WebClientOptions options = htmlUnitDriver.getWebClient().getOptions();
Assertions.assertThat(options.isThrowExceptionOnFailingStatusCode()).isFalse();
~~~ 
So, only verify if the properties have been parsed and set to the WebClient.

Just for clarification - your test if fine - keep it as it is, just add another one for the verification of the options.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/htmlunit/DroneHtmlUnitDriver.java
```diff
@@ -0,0 +1,23 @@
+package org.jboss.arquillian.drone.webdriver.htmlunit;
+
+import com.gargoylesoftware.htmlunit.WebClient;
+import org.openqa.selenium.Capabilities;
+
+public class DroneHtmlUnitDriver extends org.openqa.selenium.htmlunit.HtmlUnitDriver {
```
dipak-pawar: :+1:

docs/configuring-drone-instances.adoc
```diff
@@ -117,6 +117,12 @@ FirefoxDriver
 |Path to Firefox user preferences. This file will be parsed and values
 will be applied to freshly created Firefox profile.
 
+|htmlUnitWebClientOptions
+|-
+|Comma separated list of webClientOptions for HtmlUnitDriver. e.g.
```
dipak-pawar: :+1:

MatousJobanek:
LGTM 👍 
Good job @dipak-pawar 🥇 🚀 
Thanks

PR #88, opened 2017-05-10
ARQ-2112 As the last PhantomJS release is considered 2.1.1 so Drone won't check if there is any newer one

MatousJobanek:
The reason is that there was announced the end of the development of PhantomJS:
https://groups.google.com/forum/#!msg/phantomjs/9aI5d-LDuNE/5Z3SMZrqAQAJ
so it is expected that there won't be any newer version than the last one: 2.1.1

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
@@ -1,7 +1,5 @@
 package org.jboss.arquillian.drone.webdriver.binary.downloading.source;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonElement;
 import org.jboss.arquillian.drone.webdriver.binary.downloading.ExternalBinary;
 import org.jboss.arquillian.drone.webdriver.utils.GitHubLastUpdateCache;
 import org.jboss.arquillian.drone.webdriver.utils.HttpClient;
@@ -20,32 +18,44 @@ public class PhantomJSGitHubBitbucketSource extends GitHubSource {
 
     private static String BASE_DOWNLOAD_URL = "https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-";
 
+    private static String lastPhantomJSRelease = "2.1.1";
+
     public PhantomJSGitHubBitbucketSource(HttpClient httpClient, GitHubLastUpdateCache gitHubLastUpdateCache) {
         super("ariya", "phantomjs", httpClient, gitHubLastUpdateCache);
     }
 
-    @Override
+    /**
+     * As there was announced the end of the development of PhantomJS:
+     * https://groups.google.com/forum/#!msg/phantomjs/9aI5d-LDuNE/5Z3SMZrqAQAJ
+     * it is not necessary to check the latest release - it is expected that there won't be any newer than the last
+     * one: 2.1.1
+     * If the development of PhantomJS is resurrected, then the logic will be uncommented.
+     */
     public ExternalBinary getLatestRelease() throws Exception {
+        // See the javadoc of this method for explanation why this code is commented out
 
-        final HttpClient.Response response =
-            sentGetRequestWithPagination(getProjectUrl() + TAGS_URL, 1, lastModificationHeader());
-        final ExternalBinary latestPhantomJSBinary;
+        //        final HttpClient.Response response =
+        //            sentGetRequestWithPagination(getProjectUrl() + TAGS_URL, 1, lastModificationHeader());
+        //        final ExternalBinary latestPhantomJSBinary;
+        //
+        //        if (response.hasPayload()) {
+        //            JsonArray releaseTags = getGson().fromJson(response.getPayload(), JsonElement.class).getAsJsonArray();
+        //            if (releaseTags.size() == 0) {
+        //                return null;
+        //            }
+        //            String version = releaseTags.get(0).getAsJsonObject().get(TAG_NAME).getAsString();
+        //            latestPhantomJSBinary = new ExternalBinary(version);
+        //
+        //            latestPhantomJSBinary.setUrl(getUrlForVersion(version));
+        //            getCache().store(latestPhantomJSBinary, getUniqueKey(), extractModificationDate(response));
+        //        } else {
+        //            latestPhantomJSBinary = getCache().load(getUniqueKey(), ExternalBinary.class);
+        //        }
 
-        if (response.hasPayload()) {
-            JsonArray releaseTags = getGson().fromJson(response.getPayload(), JsonElement.class).getAsJsonArray();
-            if (releaseTags.size() == 0) {
-                return null;
-            }
-            String version = releaseTags.get(0).getAsJsonObject().get(TAG_NAME).getAsString();
-            latestPhantomJSBinary = new ExternalBinary(version);
+        ExternalBinary lastPhantomJSVersion = new ExternalBinary(lastPhantomJSRelease);
+        lastPhantomJSVersion.setUrl(getUrlForVersion(lastPhantomJSRelease));
 
-            latestPhantomJSBinary.setUrl(getUrlForVersion(version));
-            getCache().store(latestPhantomJSBinary, getUniqueKey(), extractModificationDate(response));
-        } else {
-            latestPhantomJSBinary = getCache().load(getUniqueKey(), ExternalBinary.class);
-        }
-
-        return latestPhantomJSBinary;
+        return lastPhantomJSVersion;
     }
 
     @Override
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
@@ -27,7 +27,8 @@ public class PhantomJSGitHubBitbucketSource extends GitHubSource {
      * it is not necessary to check the latest release - it is expected that there won't be any newer than the last
      * one: 2.1.1
      * If the development of PhantomJS is resurrected, then the original logic will be brought back. In this case
-     * as a reference use this commit: 5f4f64146dfbb42b641464dfb89aaa811b008a31
+     * as a reference use this logic:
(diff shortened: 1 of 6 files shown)


lordofthejars:
LGTM

bartoszmajsak:
Are we sure it won't be forked or maintained by someone else?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
```diff
@@ -20,32 +18,44 @@
...
+        //                return null;
+        //            }
+        //            String version = releaseTags.get(0).getAsJsonObject().get(TAG_NAME).getAsString();
+        //            latestPhantomJSBinary = new ExternalBinary(version);
+        //
+        //            latestPhantomJSBinary.setUrl(getUrlForVersion(version));
+        //            getCache().store(latestPhantomJSBinary, getUniqueKey(), extractModificationDate(response));
+        //        } else {
+        //            latestPhantomJSBinary = getCache().load(getUniqueKey(), ExternalBinary.class);
+        //        }
```
bartoszmajsak: Instead of leaving dead code just refer to the commit where it was implemented?

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSourceLatestReleaseFromTagsTestCase.java
```diff
@@ -30,6 +32,8 @@
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
+@Ignore("For the evaluation the PhantomJSGitHubBitbucketSource is used. But it doesn't use this logic any more."
```
bartoszmajsak: Why we ignore this? Maybe we use other repo instead? This test still has value.

MatousJobanek:
Of course that I'm not sure. This is also the reason why I commented the code out and not removed. if the development is resurrected then I will uncomment it. 
In addition, the release frequency has been very very low, so I'm thinking if it made sense to implemented this check.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSourceLatestReleaseFromTagsTestCase.java
bartoszmajsak: We still need this logic for gecko, right?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
MatousJobanek: Yeah, I wasn't sure which approach I should choose - I was thinking about both of them...

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
```diff
@@ -20,32 +18,44 @@
 
     private static String BASE_DOWNLOAD_URL = "https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-";
 
+    private static String lastPhantomJSRelease = "2.1.1";
+
     public PhantomJSGitHubBitbucketSource(HttpClient httpClient, GitHubLastUpdateCache gitHubLastUpdateCache) {
```
bartoszmajsak: PhantomJSGitHubBitbucketSource? So Bitbucket or GitHub? I'm confused :)

MatousJobanek: :-D both - from GH the tag is retrieved but the zip files are downloaded from Bitbucket

bartoszmajsak: This ain't good? https://github.com/ariya/phantomjs/releases

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSourceLatestReleaseFromTagsTestCase.java
MatousJobanek: For gecko there is: https://github.com/arquillian/arquillian-extension-drone/blob/master/drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSourceLatestReleaseTestCase.java

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
MatousJobanek: I don't follow now...  there are only tags - you don't have an access/link to the zip files

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSourceLatestReleaseFromTagsTestCase.java
bartoszmajsak: So what's the value of this test?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
bartoszmajsak: What about https://github.com/ariya/phantomjs/archive/2.1.0.zip ?

MatousJobanek: Correct me if I'm wrong - the zip contains "only" source code, not specific binaries: https://bitbucket.org/ariya/phantomjs/downloads/

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
MatousJobanek: Fixed

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
MatousJobanek: how about this: https://github.com/arquillian/arquillian-extension-drone/pull/88/commits/fc78f3c01ef916b13e3a1ade15b66c491e2fbe2c#diff-00a1f71655dc18c79ab7c7f8c0735aa3R180

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSourceLatestReleaseFromTagsTestCase.java
MatousJobanek: how about this: https://github.com/arquillian/arquillian-extension-drone/pull/88/commits/fc78f3c01ef916b13e3a1ade15b66c491e2fbe2c#diff-00a1f71655dc18c79ab7c7f8c0735aa3R180

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
bartoszmajsak: Correct :)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
```diff
@@ -15,37 +13,29 @@
...
         super("ariya", "phantomjs", httpClient, gitHubLastUpdateCache);
     }
 
-    @Override
+    /**
+     * As there was announced the end of the development of PhantomJS:
+     * https://groups.google.com/forum/#!msg/phantomjs/9aI5d-LDuNE/5Z3SMZrqAQAJ
+     * it is not necessary to check the latest release - it is expected that there won't be any newer than the last
+     * one: 2.1.1
+     * If the development of PhantomJS is resurrected, then the original logic will be brought back. In this case
```
bartoszmajsak: I would suggest:

```
* If the development of PhantomJS is resurrected, then the original logic will be brought back. In this case
* as a reference use this logic: https://github.com/arquillian/arquillian-extension-drone/blob/5f4f64146dfbb42b641464dfb89aaa811b008a31/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java#L30-L51

```

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
bartoszmajsak: Maybe we can comment why is it called that way in class doc?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
MatousJobanek: OK, done

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/PhantomJSGitHubBitbucketSource.java
MatousJobanek: Currently, there is: 
~~~ java
* A slightly changed {@link GitHubSource} implementation handling PhantomJS binaries. The latest version is retrieved
* from list of GH tags and download URL is constructed to use Bitbucket downloads storage.
~~~
so I thought that it is enough

PR #89, opened 2017-05-18
ARQ-2114 It was easy to reach the GH rate limit when firefox is used in large test suites running in CI

MatousJobanek:
- When a request with the first page is sent, then the parameter is not added -the reason is that conditional requests don't seem to be supported for parametrized requests

- When GH release is retrieved for a version then it check for the presence every page separately (lazily)

- When version is used, check for previous downloads is performed before getting the download URL

- Added logging about sending requests

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/HttpClient.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/HttpClient.java
@@ -1,9 +1,5 @@
 package org.jboss.arquillian.drone.webdriver.utils;
 
-import java.io.IOException;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
@@ -12,7 +8,16 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.util.EntityUtils;
 
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Logger;
+
 public class HttpClient {
+
+    private final Logger log = Logger.getLogger(HttpClient.class.getName());
 
     public Response get(String url) throws IOException {
         return get(url, Collections.emptyMap());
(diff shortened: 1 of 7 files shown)


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/HttpClient.java
```diff
@@ -22,6 +27,7 @@ public Response get(String url, Map<String, String> headers) throws IOException
         try (CloseableHttpClient client = createHttpClient()) {
             final HttpGet request = new HttpGet(url);
             addHeaders(headers, request);
+            log.info("Sending request: " + request + " with headers: " + Arrays.asList(request.getAllHeaders()));
```
bartoszmajsak: Do we really need to make it `info` level?

bartoszmajsak:
Can we also think about letting users refer to their GH tokens to be used? This would solve a problem of any other potential GH calls we will use in the future.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/HttpClient.java
MatousJobanek: I'm not sure about this. WDYT? which level would be more appropriate? Debug? Fine?

MatousJobanek:
Yeah, it would solve the problem. I was always trying to avoid it - not to use the credentials in CI.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/HttpClient.java
bartoszmajsak: Personally I am not interested to see if requests are sent. Only when something is not working I would be interested in seeing this information - so maybe debug with a simple way to enable it?

MatousJobanek:
In this case, the main problem was when some specific version was used - I forgot to check the directory as the first thing - so it led to the state when two requests were sent in every run.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/HttpClient.java
MatousJobanek: > simple way to enable it?

:-D this is the reason why I didn't want to use a lower level... but I could use debug level or INFO when -Darquillian.debug is used

PR #90, opened 2017-05-24
improvement in checks in automatic download feature

MatousJobanek:
There are 2 things that have been improved/changed:
* When Drone checks the zip if is already downloaded it takes into account the expected name of the file (based on version and platform)
* GitHub binaries - IAE is thrown for not found versions and a list of available versions is appended to the exception message.

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
@@ -56,15 +56,6 @@ public abstract class GitHubSource implements ExternalBinarySource {
         this.cache = gitHubLastUpdateCache;
     }
 
-    /**
-     * It is expected that this abstract method should return a regex that represents an expected file name
-     * of the release asset. These names are visible on pages of some specific release or accessible
-     * via api.github request.
-     *
-     * @return A regex that represents an expected file name of an asset associated with the required release.
-     */
-    protected abstract String getExpectedFileNameRegex(String version);
-
     @Override
     public ExternalBinary getLatestRelease() throws Exception {
         final HttpClient.Response response =
@@ -140,7 +131,7 @@ public abstract class GitHubSource implements ExternalBinarySource {
         for (JsonElement asset : assets) {
             JsonObject assetJson = asset.getAsJsonObject();
             String name = assetJson.get(assetNameKey).getAsString();
-            if (name.matches(getExpectedFileNameRegex(version))) {
+            if (name.matches(getFileNameRegexToDownload(version))) {
                 return assetJson.get(browserDownloadUrlKey).getAsString();
             }
         }
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
@@ -107,9 +107,8 @@ public abstract class GitHubSource implements ExternalBinarySource {
                 break;
             }
         }
-        log.warning(
+        throw new IllegalArgumentException(
             "There wasn't found any release for the version: " + version + " in the repository: " + projectUrl);
-        return null;
     }
 
     private ExternalBinary getReleaseForVersion(String version, JsonArray releases) throws Exception {
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
@@ -13,8 +13,10 @@ import org.jboss.arquillian.drone.webdriver.utils.Rfc2126DateTimeFormatter;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
 
@@ -92,13 +94,15 @@ public abstract class GitHubSource implements ExternalBinarySource {
     public ExternalBinary getReleaseForVersion(String version) throws Exception {
         String url = projectUrl + RELEASES_URL;
         int pageNumber = 1;
+        List<String> availableVersions = new ArrayList<>();
 
         while (true) {
             HttpClient.Response response = sentGetRequestWithPagination(url, pageNumber++, Collections.emptyMap());
             JsonElement releases = gson.fromJson(response.getPayload(), JsonElement.class);
 
             if (releases != null && releases.isJsonArray() && releases.getAsJsonArray().size() > 0) {
-                ExternalBinary releaseForVersion = getReleaseForVersion(version, releases.getAsJsonArray());
+                ExternalBinary releaseForVersion =
+                    getReleaseForVersion(version, releases.getAsJsonArray(), availableVersions);
 
                 if (releaseForVersion != null) {
                     return releaseForVersion;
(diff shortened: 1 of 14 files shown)


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
```diff
@@ -116,12 +111,13 @@ public ExternalBinary getReleaseForVersion(String version) throws Exception {
                 break;
             }
         }
-        log.warning(
-            "There wasn't found any release for the version: " + version + " in the repository: " + projectUrl);
-        return null;
+        throw new IllegalArgumentException(
+            "There wasn't found any release for the version: " + version + " in the repository: " + projectUrl
```
bartoszmajsak: How about "No release matching version xxx has been found in the repository yy......" ?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
```diff
@@ -101,13 +94,15 @@ protected ZonedDateTime extractModificationDate(HttpClient.Response response) {
...
         int pageNumber = 1;
+        List<String> availableVersions = new ArrayList<>();
 
         while (true) {
             HttpClient.Response response = sentGetRequestWithPagination(url, pageNumber++, Collections.emptyMap());
             JsonElement releases = gson.fromJson(response.getPayload(), JsonElement.class);
 
             if (releases != null && releases.isJsonArray() && releases.getAsJsonArray().size() > 0) {
-                ExternalBinary releaseForVersion = getReleaseForVersion(version, releases.getAsJsonArray());
+                ExternalBinary releaseForVersion =
```
bartoszmajsak: Just wondering if we could improve somehow this `while (true)` flow?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/EdgeDriverBinaryHandler.java
```diff
@@ -108,6 +109,11 @@ public ExternalBinary getReleaseForVersion(String version) throws Exception {
             return new ExternalBinary(driverVersion, webDriverUrl);
         }
 
+        @Override
+        public String getFileNameRegexToDownload(String version) {
+            return null;
```
bartoszmajsak: Is it ok to return `null`? Looks a bit odd.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/LocalBinarySource.java
```diff
@@ -40,4 +41,9 @@ public ExternalBinary getReleaseForVersion(String version) throws Exception {
             return getLatestRelease();
         }
     }
+
+    @Override
+    public String getFileNameRegexToDownload(String version) {
+        return null;
```
bartoszmajsak: What does it mean when we return `null`?

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/handler/BinaryHandlerDesiredVersionTestCase.java
```diff
@@ -0,0 +1,115 @@
...
+        @Override
+        public ExternalBinary getLatestRelease() throws Exception {
+            return null;
+        }
+
+        @Override
+        public ExternalBinary getReleaseForVersion(String version) throws Exception {
+            if (version.equals(FILE_VERSION_SHOULD_EXIST)){
+                Assertions.fail("The method ExternalBinarySource#getReleaseForVersion should not have been called"
+                                    + " as the file should be present in the downloaded directory and should be matched");
```
bartoszmajsak: We can also think about using mocks for that and certain tests will just ensure that this method was never called. WDYT?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
MatousJobanek: Yeah, why not :-)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/EdgeDriverBinaryHandler.java
MatousJobanek: not in this case - I forgot to change this... good catch! thanks

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/LocalBinarySource.java
MatousJobanek: Well, in this case (test) it doesn't matter, but I agree that `null` may look add - the regex `".*"` could be more convenient

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/handler/BinaryHandlerDesiredVersionTestCase.java
MatousJobanek: Yes, this could work too. I'll try to change it

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/EdgeDriverBinaryHandler.java
MatousJobanek: Actually, looking at the code again - it is OK - here https://github.com/arquillian/arquillian-extension-drone/pull/90/files#diff-71c8ade0188a564bdff67fcb52737394R181 I check if the regex is null or not - if it is null, then I don't filter the directory for any files... The reason is that in the case of "Edge" I don't know the name of the file that will be downloaded.

MatousJobanek: I'll add additional javadoc or comment explaining why it is null...

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
MatousJobanek: Yes, something like this should work too:
~~~java
 @Override
    public ExternalBinary getReleaseForVersion(String version) throws Exception {
        String url = projectUrl + RELEASES_URL;
        int pageNumber = 1;
        List<String> availableVersions = new ArrayList<>();
        JsonElement releases = getReleases(url, pageNumber);

        while (releases != null && releases.isJsonArray() && releases.getAsJsonArray().size() > 0) {

            ExternalBinary releaseForVersion =
                getReleaseForVersion(version, releases.getAsJsonArray(), availableVersions);

            if (releaseForVersion != null) {
                return releaseForVersion;
            }
            releases = getReleases(url, ++pageNumber);
        }

        throw new IllegalArgumentException(
            "There wasn't found any release for the version: " + version + " in the repository: " + projectUrl
                + " Available versions are: " + availableVersions);
    }

    private JsonElement getReleases(String url, int pageNumber) throws Exception {
        HttpClient.Response response = sentGetRequestWithPagination(url, pageNumber, Collections.emptyMap());
        return gson.fromJson(response.getPayload(), JsonElement.class);
    }
~~~

bartoszmajsak: We can also extract the condition in `while` to a method with some helpful name.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/handler/BinaryHandlerDesiredVersionTestCase.java
bartoszmajsak: Leave it for now, but please make a task out of it so we won't forget.

MatousJobanek: already changed - it was easy :-)

MatousJobanek: but I know - the annotation :-) I'll remove it in a sec

bartoszmajsak:
triumph-colbert-report

Issue #91, opened 2017-05-26
[Drone] Add chrome-headless browser option to use headless chromium


hemanik:
##### Issue Overview

Add support for chrome-headless browser option to use headless chromium

##### Expected Behaviour
The idea of this task is making it simpler - just using `chrome-headless` as a browser property:

```
<property name="browser">chrome-headless</property>
```

##### Current Behaviour
Currently, the only way to use headless chromium is to use chrome as a browser and then add `--headless` as a `chromeArguments` property:

```
<property name="browser">chrome</property>
<property name="chromeArguments">--headless</property>
```

PR #92, opened 2017-05-26
Adds chrome headless as Browser property for Drone 

hemanik:
Currently, the only way to use headless chromium is to use chrome as a browser and then add 
`--headless` as a `chromeArguments` property:
```
<property name="browser">chrome</property>
<property name="chromeArguments">--headless</property>
```
The idea of this task is making it simpler - just using chromeHeadless as a browser property:
```
<property name="browser">chromeHeadless</property>
```

- Adds support for chrome headless browser as a browser property.
- Adds test to verify ChromeHeadlessDriver for a simple loginPage.

**Fixes**: #91

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/BrowserCapabilitiesList.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/BrowserCapabilitiesList.java
@@ -225,7 +225,7 @@ public class BrowserCapabilitiesList {
     public static class chromeHeadless implements BrowserCapabilities {
         @Override
         public String getReadableName() {
-            return "chromeHeadless";
+            return "chromeheadless";
         }
 
         @Override
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/BrowserCapabilitiesList.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/BrowserCapabilitiesList.java
@@ -130,8 +130,6 @@ public class BrowserCapabilitiesList {
         }
     }
 
-    ;
-
     public static class Opera implements BrowserCapabilities {
 
         @Override
@@ -154,8 +152,6 @@ public class BrowserCapabilitiesList {
             return 0;
         }
     }
-
-    ;
 
     public static class Remote implements BrowserCapabilities {
 
@@ -180,8 +176,6 @@ public class BrowserCapabilitiesList {
         }
     }
 
-    ;
-
     public static class Safari implements BrowserCapabilities {
 
         @Override
@@ -204,8 +198,6 @@ public class BrowserCapabilitiesList {
             return 0;
         }
     }
-
-    ;
 
     public static class PhantomJS implements BrowserCapabilities {
 
@@ -230,5 +222,25 @@ public class BrowserCapabilitiesList {
         }
     }
 
-    ;
+    public static class chromeHeadless implements BrowserCapabilities {
+        @Override
+        public String getReadableName() {
+            return "chromeHeadless";
+        }
+
+        @Override
+        public String getImplementationClassName() {
+            return "org.openqa.selenium.chrome.ChromeDriver";
+        }
+
+        @Override
+        public Map<String, ?> getRawCapabilities() {
+            return DesiredCapabilities.chrome().asMap();
+        }
+
+        @Override
+        public int getPrecedence() {
+            return 0;
+        }
+    }
 }
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
@@ -114,7 +114,7 @@ public class ChromeDriverFactory extends AbstractWebDriverFactory<ChromeDriver> 
         return capabilities;
     }
 
-    public static void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabilities capabilities) {
+    public void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabilities capabilities) {
         ChromeOptions chromeOptions = new ChromeOptions();
 
         String browser = configuration.getBrowser().toLowerCase();
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
@@ -102,21 +102,30 @@ public class ChromeDriverFactory extends AbstractWebDriverFactory<ChromeDriver> 
 
         new ChromeDriverBinaryHandler(capabilities).checkAndSetBinary(performValidations);
 
-        ChromeOptions chromeOptions = new ChromeOptions();
-
-        String browser = configuration.getBrowser();
-        if (browser.equals("chromeHeadless") || browser.equals("chromeheadless")) {
-            chromeOptions.addArguments("--headless");
-        }
-
-        CapabilitiesOptionsMapper.mapCapabilities(chromeOptions, capabilities, BROWSER_CAPABILITIES);
-
         // verify binary capabilities
         if (Validate.nonEmpty(binary)) {
             if (performValidations) {
                 Validate.isExecutable(binary, "Chrome binary must point to an executable file, " + binary);
             }
+        }
 
+        setChromeOptions(configuration, capabilities);
+
+        return capabilities;
+    }
+
+    public static void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabilities capabilities) {
+        ChromeOptions chromeOptions = new ChromeOptions();
+
+        String browser = configuration.getBrowser().toLowerCase();
+        if (browser.equals("chromeheadless")) {
+            chromeOptions.addArguments("--headless");
+        }
+
+        CapabilitiesOptionsMapper.mapCapabilities(chromeOptions, capabilities, BROWSER_CAPABILITIES);
+
+        String binary = (String) capabilities.getCapability("chrome.binary");
+        if (Validate.nonEmpty(binary)) {
             // ARQ-1823 - setting chrome binary path through ChromeOptions
             chromeOptions.setBinary(binary);
         }
@@ -136,8 +145,6 @@ public class ChromeDriverFactory extends AbstractWebDriverFactory<ChromeDriver> 
                 log.warning(e.getMessage());
             }
         }
-
-        return capabilities;
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


MatousJobanek:
@hemanik have you figured out why the chromeHeadless version is failing in case of reusable tests?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -103,6 +103,11 @@ public Capabilities getCapabilities(WebDriverConfiguration configuration, boolea
         new ChromeDriverBinaryHandler(capabilities).checkAndSetBinary(performValidations);
 
         ChromeOptions chromeOptions = new ChromeOptions();
+
+        if (capabilities.getBrowserName().equals("chromeHeadless")) {
```
MatousJobanek: Drone configuration is case sensitive so `chromeheadless` is also valid

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/factory/ChromeHeadlessDriverTest.java
```diff
@@ -0,0 +1,52 @@
...
+public class ChromeHeadlessDriverTest {
+
+    @BeforeClass
+    public static void executeOnlyIfChromeOrChromeHeadlessBrowser() {
+        final String browser = System.getProperty("browser");
+        Assume.assumeTrue(browser.equals("chrome") || browser.equals("chromeHeadless"));
+    }
+
+    @Test
+    public void testOpenSimplePageUsingHeadlessChromium() throws IOException {
```
MatousJobanek: if you name it `testOpenSimplePageUsingHeadlessChromium` then why you allow this test also for property `chrome`
in addition - it is not chromium but chrome.

MatousJobanek: The test only checks if it opens a page, it doesn't check if the argument is set and the headless version is really used... Please add other tests for that...

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/BrowserCapabilitiesList.java
```diff
@@ -230,5 +222,27 @@ public int getPrecedence() {
...
+
+        @Override
+        public String getImplementationClassName() {
+            return "org.openqa.selenium.chrome.ChromeDriver";
+        }
+
+        @Override
+        public Map<String, ?> getRawCapabilities() {
+            DesiredCapabilities headlessChrome = DesiredCapabilities.chrome();
+            headlessChrome.setBrowserName("chromeHeadless");
```
hemanik: Had to set the `browserName` here, otherwise, when DesiredCapabilities class is called internally, the browserName gets overridden somehow to `chrome`, and it no longer functions in the headless mode but in the normal way.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
hemanik: will update this.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/factory/ChromeHeadlessDriverTest.java
hemanik: Will fix this.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/BrowserCapabilitiesList.java
hemanik: Tried adding different properties and checking if chrome headless version supports reusing browser session but did not find any concrete reason for it.
There is no evidence to support that it is not does not support browser session reuse.

MatousJobanek:
> Tried adding different properties and checking if chrome headless version supports reusing browser session but did not find any concrete reason for it.
There is no evidence to support that it is not does not support browser session reuse.

Chrome itself supports session reuse, and I wouldn't expect that the argument --headless would change the behavior significantly so it wouldn't be possible to use there as well. I believe that the problem is somewhere in our configuration.
What is the difference between the normal test and the reusable one? What says the exception? Is there any other use-case that we haven't covered yet?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/BrowserCapabilitiesList.java
MatousJobanek: correct me if I'm wrong, this is `browserName` property, you check the `browser` property, which is different, so it shouldn't have any impact on your logic. 

>the browserName gets overridden somehow to `chrome`

browserName is by default set to `chrome` as you call `DesiredCapabilities.chrome()` where the property is set.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/factory/ChromeHeadlessDriverTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        WebDriver driver = chromeDriverFactory.createInstance(configuration);
+        URL page = this.getClass().getClassLoader().getResource("simple.html");
+        driver.get(page.toString());
+        Assert.assertEquals("The page title doesn't match.", "Simple Page", driver.getTitle());
+        driver.quit();
+    }
+
+    @Test
+    public void shouldSetChromeOptionsForHeadlessChromeHeadlessDriver() throws IOException {
```
MatousJobanek: This test doesn't require the browser set - it doesn't require the browser installed on PC. In other words, you can run it even when the `-Dbrowser=chromeHeadless` is not set.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/factory/remote/reusable/TestRemoteWebDriverFactorySessionStoring.java
```diff
@@ -75,8 +75,9 @@
     private InitializationParameter initializationParameter;
 
     @BeforeClass
-    public static void skipIfEdgeBrowser() {
-        Assume.assumeFalse(System.getProperty("browser", "phantomjs").equals("edge"));
+    public static void skipIfEdgeOrChromeHeadlessBrowser() {
+        String browser = System.getProperty("browser", "phantomjs");
+        Assume.assumeFalse(browser.equals("edge") || browser.equals("chromeHeadless") || browser.equals("chromeheadless"));
```
MatousJobanek: why do we skip it? in this case, it should work too... You just need check when the chromeheadless is used and in then set appropriate ChromeOption argument

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -103,6 +103,12 @@ public Capabilities getCapabilities(WebDriverConfiguration configuration, boolea
         new ChromeDriverBinaryHandler(capabilities).checkAndSetBinary(performValidations);
 
         ChromeOptions chromeOptions = new ChromeOptions();
+
+        String browser = configuration.getBrowser();
+        if (browser.equals("chromeHeadless") || browser.equals("chromeheadless")) {
```
hemanik: Don't we expect chrome to show up browser window unless it is in headless mode ? 
Also, if a user wishes to use `chrome` instead of `chromeheadless` and still be able to use headless mode that can still in done as per the original way. 
```
<property name="browser">chrome</property>
<property name="chromeArguments">--headless</property> 
```
Please correct me, if my interpretation here is wrong.

MatousJobanek: You are completely right. What I meant by my comment is that setting ChromeOptions can be done also for `chrome` - no only `chromeHeadless`. Of course, part of the logic will be check if the user has used `chromeHeadless` as a browser or not - if yes then set the appropriate argument there.
Setting ChromeOptions in the case of RemoteWebdriver was obviously forgotten.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -102,21 +102,30 @@ public Capabilities getCapabilities(WebDriverConfiguration configuration, boolea
...
+                Validate.isExecutable(binary, "Chrome binary must point to an executable file, " + binary);
+            }
+        }
+
+        setChromeOptions(configuration, capabilities);
+
+        return capabilities;
+    }
+
+    public static void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabilities capabilities) {
```
MatousJobanek: It doesn't have to be static - you can call 
`new ChromeDriverFactory().setChromeOptions(configuration, desiredCapabilities);`

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/factory/remote/reusable/TestRemoteWebDriverFactorySessionStoring.java
```diff
@@ -107,6 +108,12 @@ public void setupMocks() {
 
         initializationParameter = new InitializationParameter(hubUrl, desiredCapabilities);
 
+        String browser = System.getProperty("browser", "phantomjs").toLowerCase();
```
MatousJobanek: Why do you use the default value? The retrieved value is used only for checking if `chromeheadless` is used - then you don't care about some defaults....

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -102,15 +102,30 @@ public Capabilities getCapabilities(WebDriverConfiguration configuration, boolea
 
         new ChromeDriverBinaryHandler(capabilities).checkAndSetBinary(performValidations);
 
-        ChromeOptions chromeOptions = new ChromeOptions();
-        CapabilitiesOptionsMapper.mapCapabilities(chromeOptions, capabilities, BROWSER_CAPABILITIES);
-
         // verify binary capabilities
         if (Validate.nonEmpty(binary)) {
```
MatousJobanek: Shouldn't be this validation also part of setting ChromeOptions? Otherwise, (in the case of RemoveWebdriver) it may happen that you set non-executable binary...

hemanik: `performValidations` to check if the binary is executable is set as a separate condition with a separate if-clause. Doesn't it set the chromeOptions irrespective of whether the binary is executable or not as per the original logic ?

MatousJobanek: yeah, you are right with this - that the `setChromeOptions` should only set... then leave it as it is...

MatousJobanek:
LGTM thanks @hemanik

Issue #93, opened 2017-05-30
disable drone for the run using maven profile


punitbisht:
Hi , 

I am using cukespace https://github.com/cukespace/cukespace to run my features using Arquillian.
there is one issue I am facing with Tags and probably you guys can help me .

I have a suite of features, each features have few tags associated . Now when i use `mvn -PmyProfile clean "-Dcucumber.options=-t @myTag" test` the feature with tag "myTag" executes but other features also launch the browser and close it immediately.

This is expected since the browser is bound to the test which has the driver then. 

is there any way I can disable drone for those tests which are not matching with the tag.

MatousJobanek:
Hi,
thank you for your question. 
Actually, I'm not aware of any possibility to disable Drone in a test. The easiest way is removing the annotation `@Drone`. 
If the tag invokes only some of the test methods and the other ones are ignored, then you can move the WebDriver variable declaration to the test method as its parameter: 
~~~java
@Test
public void yourTest(@Drone  WebDriver browser) {
 ...
}
~~~ 
Then the Drone webdriver instance is created in test method scope and is instantiated only for this specific method. I don't know the implementation of cucumber, so I don't know how they manage skipping of the tests, so it may happen that the Drone instance is created anyway. Please give it a try and let me know if it works.

MatousJobanek:
Lazy initialization is a good idea, but I'm afraid that not easy to do :-/. But if you want to take care of it, I'll be more than happy for any contribution ;-).
On the other hand, having a flag such as `-Darquillian.drone.skip=true` that would skip initialization of the Drone instances would be quite easy to implement. This flag would be checked in `@Before` phase so it would be possible to modify it.
Would that work for you?

rmannibucau:
Hmm, the flag works but thought it was harder to impl than lazy init (which is basically just a kind of proxy) cause the flag requires to set it in a lot of place. But if that's easier for you +1 :), in any case it solves the root issue.

punitbisht:
Thanks @MatousJobanek  @rmannibucau  for considering this feature and making quick changes

punitbisht:
@MatousJobanek is this available with the release??

MatousJobanek:
Yes, forgot to mention it here - it is part of Drone 2.2.0

PR #94, opened 2017-06-01
ARQ-2119 Introduced system property for skippping Drone creation/injection

MatousJobanek:
The property is `arquillian.drone.skip.creation`. When it is set to `true` then a Drone instance is not created and injected.
This property is checked in a phase `@Before` so it is possible to modify the property during the test execution.
Fix for issue #93

### DIFF
--- a/drone-impl/src/test/java/org/jboss/arquillian/drone/impl/EnricherTestCase.java
+++ b/drone-impl/src/test/java/org/jboss/arquillian/drone/impl/EnricherTestCase.java
@@ -55,6 +55,8 @@ import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.junit.MockitoJUnitRunner;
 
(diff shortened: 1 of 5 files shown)


drone-impl/src/test/java/org/jboss/arquillian/drone/impl/EnricherTestCase.java
```diff
@@ -111,6 +113,11 @@ public void setMocks() {
         Mockito.when(deploymentDescription2.getName()).thenReturn(AnnotationMocks.DEPLOYMENT_2);
     }
 
+    @org.junit.After
+    public void resetProperty(){
+        System.clearProperty(ARQUILLIAN_DRONE_CREATION_PROPERTY);
```
bartoszmajsak: You might also want to look at http://stefanbirkner.github.io/system-rules/

punitbisht:
@MatousJobanek  how this property will behave when we are running test parallel with mvn fork

PR #95, opened 2017-06-02
feat: Handles response with rate limit and with other errors

MatousJobanek:
Check if the rate limit has exceeded - if yes it tries to use the cached version and log warning, otherwise throws an exception with explaining message

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
@@ -4,21 +4,21 @@ import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
-import org.apache.http.client.utils.URIBuilder;
-import org.jboss.arquillian.drone.webdriver.binary.downloading.ExternalBinary;
-import org.jboss.arquillian.drone.webdriver.utils.GitHubLastUpdateCache;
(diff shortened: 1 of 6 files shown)


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
```diff
@@ -175,4 +222,14 @@ protected String getUniqueKey() {
     protected GitHubLastUpdateCache getCache() {
         return cache;
     }
+
+    // workaround for https://github.com/SpectoLabs/hoverfly/issues/573
+    protected String getRateLimitRemainingHeaderKey() {
```
bartoszmajsak: Is it really necessary to have workarounds in the production code for the tools we use in testing?

bartoszmajsak: Does the issue still persist with `0.6.1`? https://github.com/SpectoLabs/hoverfly/issues/573#issuecomment-306159723

benjih: Hey @bartoszmajsak,

0.6.1 of hoverfly-java should have this issue fixed.

MatousJobanek: Yes, this is fixed in the release - haven't had time to upgrade it so far... Thank you @benjih

PR #98, opened 2017-06-23
Sets window-size as chrome-option for chrome headless mode.

hemanik:
- Some websites and pages are broken if loaded in a browser window that is half the total screen size of the monitor. Resizing chrome is buggy in latest chrome-drivers and explicitly setting the window size when browser is opened is required.
- Improves the execution speed for Drone Test Suite when using `chromeHeadless` browser, by eliminating waiting time for window resizing. 

- Set window size as argument via chromeOptions.

**Fixes**: #

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
@@ -19,6 +19,7 @@ package org.jboss.arquillian.drone.webdriver.factory;
 import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
 import org.jboss.arquillian.config.descriptor.api.ArquillianDescriptor;
 import org.jboss.arquillian.drone.spi.Configurator;
 import org.jboss.arquillian.drone.spi.Destructor;
@@ -31,6 +32,8 @@ import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.chrome.ChromeDriver;
 import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.remote.DesiredCapabilities;
+
+import static org.jboss.arquillian.drone.webdriver.window.WindowResizer.DIMENSIONS_PATTERN;
 
 /**
  * Factory which combines {@link org.jboss.arquillian.drone.spi.Configurator},
@@ -120,6 +123,17 @@ public class ChromeDriverFactory extends AbstractWebDriverFactory<ChromeDriver> 
         String browser = configuration.getBrowser().toLowerCase();
         if (browser.equals("chromeheadless")) {
             chromeOptions.addArguments("--headless");
+            if (configuration.getDimensions() != null) {
+                String dimensions = configuration.getDimensions().toLowerCase().trim();
+                Matcher m = DIMENSIONS_PATTERN.matcher(dimensions);
+                if (m.matches()) {
+                    String width = m.group(1);
+                    String height = m.group(2);
+                    chromeOptions.addArguments(String.format("--window-size=%s,%s", width, height));
+                } else if (dimensions.equals("full") || dimensions.equals("fullscreen") || dimensions.equals("max")) {
+                    chromeOptions.addArguments("--window-size=1920,1080"); // workaround till a better way is found.
+                }
+            }
         }
 
         CapabilitiesOptionsMapper.mapCapabilities(chromeOptions, capabilities, BROWSER_CAPABILITIES);
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
@@ -19,7 +19,6 @@ package org.jboss.arquillian.drone.webdriver.factory;
 import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.regex.Matcher;
 import org.jboss.arquillian.config.descriptor.api.ArquillianDescriptor;
 import org.jboss.arquillian.drone.spi.Configurator;
 import org.jboss.arquillian.drone.spi.Destructor;
@@ -28,12 +27,11 @@ import org.jboss.arquillian.drone.spi.Instantiator;
 import org.jboss.arquillian.drone.webdriver.binary.handler.ChromeDriverBinaryHandler;
 import org.jboss.arquillian.drone.webdriver.configuration.WebDriverConfiguration;
 import org.jboss.arquillian.drone.webdriver.utils.Validate;
+import org.jboss.arquillian.drone.webdriver.window.Dimensions;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.chrome.ChromeDriver;
 import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.remote.DesiredCapabilities;
-
-import static org.jboss.arquillian.drone.webdriver.window.WindowResizer.DIMENSIONS_PATTERN;
 
 /**
  * Factory which combines {@link org.jboss.arquillian.drone.spi.Configurator},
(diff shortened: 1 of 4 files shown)


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -120,6 +120,7 @@ public void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabi
         String browser = configuration.getBrowser().toLowerCase();
         if (browser.equals("chromeheadless")) {
             chromeOptions.addArguments("--headless");
+            chromeOptions.addArguments("--window-size=1200,600");
```
MatousJobanek: I'm not sure if setting specific sizes is the best approach - why this size and not bigger or smaller one...? I'm not also sure if this (or other) sizes wouldn't cause unexpected problems on some platforms/PC configurations. 
I can imagine that we could use the size specified in `dimensions` property, otherwise, I wouldn't set there anything.

hemanik: The `size=1200x600` was an approximation for a near about fullscreen size. Couldn't come up with the best size(may be smaller or larger). 
`dimensions` property didn't work, when executing test-suite tests with chromeheadless, it gave the error msg `could not automatically resize browser.` 
Also, when executing drone tests after this change, removed the warning `waiting for browser to resize`.

MatousJobanek: I know that the `dimension` doesn't work for this case - because there is nothing implemented like this... 
What I meant is adding the support for it - that when somebody uses chromeHeadless browser and specifies dimensions, then the `--window-size` argument should be set automatically.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -120,6 +123,17 @@ public void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabi
         String browser = configuration.getBrowser().toLowerCase();
         if (browser.equals("chromeheadless")) {
             chromeOptions.addArguments("--headless");
+            if (configuration.getDimensions() != null) {
+                String dimensions = configuration.getDimensions().toLowerCase().trim();
+                Matcher m = DIMENSIONS_PATTERN.matcher(dimensions);
+                if (m.matches()) {
+                    String width = m.group(1);
+                    String height = m.group(2);
```
MatousJobanek: the lines 126 - 131 are same as those ones in the windows resizer class - 75-81. Please extract it to not have it duplicated

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -120,6 +123,17 @@ public void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabi
...
+                    chromeOptions.addArguments(String.format("--window-size=%s,%s", width, height));
+                } else if (dimensions.equals("full") || dimensions.equals("fullscreen") || dimensions.equals("max")) {
+                    chromeOptions.addArguments("--window-size=1920,1080"); // workaround till a better way is found.
```
MatousJobanek: just log warning that it si not possible to set fullscreen for headless, that user has to define specific dimensions.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/window/WindowResizer.java
```diff
@@ -70,17 +70,19 @@ private void resizeWindow(DroneEvent event) {
         WebDriverConfiguration configuration = context.get(dronePoint).getConfigurationAs(WebDriverConfiguration.class);
         Validate.stateNotNull(configuration, "WebDriver configuration must not be null");
 
-        if (configuration.getDimensions() != null) {
-            String dimensions = configuration.getDimensions().toLowerCase().trim();
-            Matcher m = DIMENSIONS_PATTERN.matcher(dimensions);
+        String browser = configuration.getBrowser().toLowerCase();
+        if(!browser.equals("chromeheadless")) {
```
MatousJobanek: btw, how it behaves in case of HtmlUnit and PhantomJS?

hemanik: Webdriver probably doesn't support automatic resizing for chrome but it does for PhantomJs and HtmlUnit. Also, seems like PhantomJs and HtmlUnit have some defaults set for the maximized mode. 
When I try executing our UI tests against these browsers. The values I get for maximized dimension are:
`PhantomJs - (1366, 768)`
`HtmlUnit -(1272, 698)`

Also, I read somewhere - 
` PhantomJS driver, by default, it will run in 400×300. You may want to run it in a desktop resolution. Don’t dispare. It is easy. You can tell Geb driver to maximize its window and it will run in  1366, 768`

MatousJobanek: Good job with this investigation. 
If both PhantomJs and HtmlUnit solves this problem using specific dimensions, then (technically) we could provide the same functionality also for Chrome headless - as you did. Then leave it as it is now - using the specific dimensions when fullscreen is used. Just log info about it...

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
MatousJobanek: See my comment below

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/window/Dimensions.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public static int getWidth() {
+        return width;
+    }
+
+    public static int getHeight() {
+        return height;
+    }
+
+    public static boolean isValidDimensionPattern(String dimensions) {
```
MatousJobanek: Don't use it as a class with static methods - the values are not intended to be used as constants. Change it to normal object class with constructor and non-static methods

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/window/DimensionsPatternTest.java
```diff
@@ -24,15 +24,15 @@
 
     @Test
     public void validSample() {
-        Matcher m = WindowResizer.DIMENSIONS_PATTERN.matcher("100x200");
+        Matcher m = Dimensions.DIMENSIONS_PATTERN.matcher("100x200");
```
MatousJobanek: in this test, you can use the object `Dimensions` to verify no only the pattern, but also the rest of the class logic

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/window/Dimensions.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public int getHeight() {
+        return height;
+    }
+
+    public boolean hasValidDimensionPattern() {
+        Matcher m = DIMENSIONS_PATTERN.matcher(dimensions);
+        if (m.matches()) {
+            setDimensions(m);
```
MatousJobanek: This means that the dimensions variables (`width` and `height`) are set only if the method `hasValidDimensionPattern` is called (this doesn't seem to be very intuitive). I would expect that the variables are tried to be set in the constructor. If it is not valid, then the values are zero. This would lead to the state when in the if statement https://github.com/arquillian/arquillian-extension-drone/pull/98/files#diff-ff926f782beea29de25d18e97da425daR127 you could check if the variables are set or not.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -120,6 +121,17 @@ public void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabi
         String browser = configuration.getBrowser().toLowerCase();
         if (browser.equals("chromeheadless")) {
             chromeOptions.addArguments("--headless");
+            if (configuration.getDimensions() != null) {
+                String windowSize = configuration.getDimensions().toLowerCase().trim();
+                Dimensions dimensions = new Dimensions(windowSize);
+                if (dimensions.hasValidDimensionPattern()) {
+                    chromeOptions.addArguments(
+                        String.format("--window-size=%d,%d", dimensions.getWidth(), dimensions.getHeight()));
+                } else if (windowSize.equals("full") || windowSize.equals("fullscreen") || windowSize.equals("max")) {
```
MatousJobanek: This check can be also part of the `Dimensions` class logic

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/ChromeDriverFactory.java
```diff
@@ -120,6 +121,17 @@ public void setChromeOptions(WebDriverConfiguration configuration, DesiredCapabi
         String browser = configuration.getBrowser().toLowerCase();
         if (browser.equals("chromeheadless")) {
             chromeOptions.addArguments("--headless");
+            if (configuration.getDimensions() != null) {
+                String windowSize = configuration.getDimensions().toLowerCase().trim();
```
MatousJobanek: This can be moved to `Dimensions` class

PR #101, opened 2017-07-04
ARQ-2038 Create Drone extension for Appium

vmuzikar:
Adds the extension for Appium.
https://issues.jboss.org/browse/ARQ-2038

- Adds the extension
- Allows mapping `ChromeOptions` from outside the package (i.e. to allow the extension to use the mapping mechanics)

**Fixes**: #

With regards to tests. Since full integration tests would require specifically configured environment with Appium server, some mobile virtual device etc., I've created just some unit-like tests.

However, I wasn't sure how to properly test the `instantiationTimeoutInSeconds`, so I "took inspiration" (i.e. basically copied [GlobalDroneConfigurationTestCase](https://github.com/arquillian/arquillian-extension-drone/blob/master/drone-impl/src/test/java/org/jboss/arquillian/drone/impl/GlobalDroneConfigurationTestCase.java)).
Please let me know if it's ok - if not, I'll try to write it differently.

### DIFF
--- a/extension/arquillian-drone-appium-extension/src/main/java/org/arquillian/drone/appium/extension/DefaultValuesModifier.java
+++ b/extension/arquillian-drone-appium-extension/src/main/java/org/arquillian/drone/appium/extension/DefaultValuesModifier.java
@@ -45,9 +45,11 @@ public class DefaultValuesModifier {
      */
     public void modifyDefaultTimeout(@Observes(precedence = -100) BeforeSuite event) {
         if (arquillianDescriptor.get()
-            .extension("drone")
+            .extension(GlobalDroneConfiguration.CONFIGURATION_NAME)
             .getExtensionProperties()
-            .get("instantiationTimeoutInSeconds") != null) return;
+            .get("instantiationTimeoutInSeconds") != null) {
+            return;
+        }
 
         GlobalDroneConfiguration globalDroneConfiguration =
             droneContext.get().getGlobalDroneConfiguration(GlobalDroneConfiguration.class);
--- a/extension/arquillian-drone-appium-extension/src/main/java/org/arquillian/drone/appium/extension/DefaultValuesModifier.java
+++ b/extension/arquillian-drone-appium-extension/src/main/java/org/arquillian/drone/appium/extension/DefaultValuesModifier.java
@@ -1,0 +1,56 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat Middleware LLC, and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.arquillian.drone.appium.extension;
+
+import org.jboss.arquillian.config.descriptor.api.ArquillianDescriptor;
+import org.jboss.arquillian.core.api.Instance;
+import org.jboss.arquillian.core.api.annotation.Inject;
+import org.jboss.arquillian.core.api.annotation.Observes;
+import org.jboss.arquillian.drone.impl.DroneLifecycleManager.GlobalDroneConfiguration;
+import org.jboss.arquillian.drone.spi.DroneContext;
+import org.jboss.arquillian.test.spi.event.suite.BeforeSuite;
+
+/**
+ * Modifies default WebDriver/Drone properties
+ *
+ * @author Vaclav Muzikar <vmuzikar@redhat.com>
+ */
+public class DefaultValuesModifier {
+    public static final int DEFAULT_INSTANTIATION_TIMEOUT = 240;
+
+    @Inject
+    private Instance<DroneContext> droneContext;
+
+    @Inject
+    private Instance<ArquillianDescriptor> arquillianDescriptor;
+
+    /**
+     * Sets the default instantiationTimeoutInSeconds Drone property
+     * @param event
+     */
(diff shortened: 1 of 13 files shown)


lordofthejars:
Thank you very much for this PR, one thing, do you think it would be possible to create a functional test, (ignored by default) that shows how to use it? Maybe using Docker + Arquillian Cube (https://hub.docker.com/r/appium/appium-emulator/) If you see that it will require some time, just an ftest to see how a test should be done, would it be enough.

Alex

lordofthejars:
PR looks good to me,.

extension/arquillian-drone-appium-extension/src/main/java/org/arquillian/drone/appium/extension/DefaultValuesModifier.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Sets the default instantiationTimeoutInSeconds Drone property
+     * @param event
+     */
+    public void modifyDefaultTimeout(@Observes(precedence = -100) BeforeSuite event) {
+        if (arquillianDescriptor.get()
+            .extension("drone")
+            .getExtensionProperties()
+            .get("instantiationTimeoutInSeconds") != null) return;
```
MatousJobanek: Use curly brackets for all if statements to keep the code consistent across the whole project

extension/arquillian-drone-appium-extension/src/main/java/org/arquillian/drone/appium/extension/DefaultValuesModifier.java
```diff
@@ -0,0 +1,56 @@
...
+    @Inject
+    private Instance<ArquillianDescriptor> arquillianDescriptor;
+
+    /**
+     * Sets the default instantiationTimeoutInSeconds Drone property
+     * @param event
+     */
+    public void modifyDefaultTimeout(@Observes(precedence = -100) BeforeSuite event) {
+        if (arquillianDescriptor.get()
+            .extension("drone")
```
MatousJobanek: instead of using the string you can use the constant CONFIGURATION_NAME https://github.com/arquillian/arquillian-extension-drone/blob/master/drone-impl/src/main/java/org/jboss/arquillian/drone/impl/DroneLifecycleManager.java#L225

extension/arquillian-drone-appium-extension/src/main/java/org/arquillian/drone/appium/extension/webdriver/AppiumDriverFactory.java
```diff
@@ -0,0 +1,142 @@
...
+        if (StringUtils.isBlank(platform)) {
+            throw new IllegalArgumentException("You have to specify " + MobileCapabilityType.PLATFORM_NAME);
+        }
+        platform = platform.toLowerCase();
+
+        Class<? extends AppiumDriver> driverClass;
+             if (MobilePlatform.ANDROID.toLowerCase().equals(platform)) driverClass = AndroidDriver.class;
+        else if (MobilePlatform.IOS.toLowerCase().equals(platform))     driverClass = IOSDriver.class;
+        else if (MobilePlatform.WINDOWS.toLowerCase().equals(platform)) driverClass = WindowsDriver.class;
+        else                                                            driverClass = AppiumDriver.class;
```
MatousJobanek: reformat to use curly brackets format. Or use the switch statement if possible.

extension/arquillian-drone-appium-extension/src/main/java/org/arquillian/drone/appium/extension/webdriver/AppiumDriverFactory.java
```diff
@@ -0,0 +1,142 @@
...
+     *     the configuration object for the extension
+     *
+     * @return The Appium WebDriver
+     */
+    @Override
+    public AppiumDriver createInstance(WebDriverConfiguration configuration) {
+        Capabilities capabilities = getCapabilities(configuration);
+
+        String platform = (String)capabilities.getCapability(MobileCapabilityType.PLATFORM_NAME);
+        if (StringUtils.isBlank(platform)) {
```
MatousJobanek: you can use our util methods: https://github.com/arquillian/arquillian-extension-drone/blob/master/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/utils/Validate.java#L49
Then if this is the only place where you use `StringUtils` then you can remove it from dependencies.

vmuzikar:
Thanks for the feedback!
I'll look into the requested changes asap.

lordofthejars:
@MatousJobanek my idea of ftest would be not exactly for documenting or for CI (as it is disabled all the time) but if a user has any problem we can quickly uncomment the annotation and check from an example that we know it should work. This could help also for example during major release version where we might want to check if everything still operates normally. Something like we do in cube.

vmuzikar:
@lordofthejars I see your point.
However, imho the functional test itself would be more complicated than the extension itself, as it would require to use (and configure) an Android Virtual Device, which is not a quick and straightforward process. This could be of course achieved e.g. by using a [docker image of Android](https://github.com/butomo1989/docker-android) which could help us with the whole Android-installation thing, but that would add another potentially error prone layer to the test.
Of course the Android configuration could be done manually (either by using pure Android SDK or the docker image) but I'm afraid it wouldn't be really straightforward and would be more like a manual testing.

Another thing is that doing such a test would result more like in testing the Appium itself.
The extension is really, really simple - all what it does is it just instantiates the AppiumDriver.
The only extra things here are the ChromeOptions and the changed default timeout. For both of them there are tests.

If the instantiation of the AppiumDriver would be broken by some future commit, I bet much more other things (which are properly tested) would be broken too, so you'd notice that something went wrong.

What I want to say, the Appium extension does nothing special and a full-featured functional test would be imho rather an over-kill.

But I guess the decision is up to @MatousJobanek. If you agree we should do a proper functional test, I'll try to add it. :)

vmuzikar:
I've tried to make the requested changes.
Could you please review?

I've done it as a one more commit. If the changes are ok, I'll then squash the commits so only one final commit would be merged.

MatousJobanek:
@lordofthejars @vmuzikar I agree with Vaclav. The ftest would be really an overkill for this small extension. So far I'm fine with the tests that are provided in this PR and for me they are sufficient.

MatousJobanek:
I squashed the commits when merging your PR. 
Thank you @vmuzikar for your contribution - really appreciated. Good job 👍 🥇

Issue #114, opened 2017-11-06
Cannot set useAutomationExtension option for chrome


kullmanp:
##### Issue Overview

I would like to set the experimental option "useAutomationExtension" to false. 

##### Expected Behaviour
It should be possible like this:

```
  <extension qualifier="webdriver">
    <property name="chromeExperimentalOption">
    {
      "useAutomationExtension": "false"
    }
    </property>
  </extension>

```

##### Current Behaviour
I get an exception:

```
com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was STRING at path $
	at com.google.gson.Gson.fromJson(Gson.java:899)
	at com.google.gson.Gson.fromJson(Gson.java:952)
	at org.jboss.arquillian.drone.webdriver.factory.CapabilitiesOptionsMapper.handleDictionaries(CapabilitiesOptionsMapper.java:108)
	at org.jboss.arquillian.drone.webdriver.factory.CapabilitiesOptionsMapper.mapCapabilities(CapabilitiesOptionsMapper.java:86)
	at org.jboss.arquillian.drone.webdriver.factory.ChromeDriverFactory.setChromeOptions(ChromeDriverFactory.java:123)
	at org.jboss.arquillian.drone.webdriver.factory.ChromeDriverFactory.getCapabilities(ChromeDriverFactory.java:114)
	at org.jboss.arquillian.drone.webdriver.factory.ChromeDriverFactory.createInstance(ChromeDriverFactory.java:79)
	at org.jboss.arquillian.drone.webdriver.factory.ChromeDriverFactory.createInstance(ChromeDriverFactory.java:41)
	at org.jboss.arquillian.drone.webdriver.factory.WebDriverFactory.createInstance(WebDriverFactory.java:127)
	at org.jboss.arquillian.drone.webdriver.factory.WebDriverFactory.createInstance(WebDriverFactory.java:38)
	at org.jboss.arquillian.drone.impl.DroneConfigurator$1.createInstance(DroneConfigurator.java:112)
	at org.jboss.arquillian.drone.impl.CachingCallableImpl.call(CachingCallableImpl.java:44)
	at org.jboss.arquillian.core.impl.threading.ThreadedExecutorService$ContextualCallable.call(ThreadedExecutorService.java:89)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #118, opened 2017-11-10
Allow typed primitive option values other than strings

kullmanp:
Unfortunately, the fix for https://github.com/arquillian/arquillian-extension-drone/issues/114 does not work with selenium. I should have specified the json primitive false instead of the string "false". 

This pull requst allows one to specify a boolean: 
```

<property name="chromeExperimentalOption">
    {
      "useAutomationExtension": false
    }
    </property>
```


MatousJobanek:
>Unfortunately, the fix for #114 does not work with selenium. 

Yeah, my bad, I should have tried to directly pass it to the selenium and see it in a bigger picture.

Thx a lot for fixing it. I'm very glad for any contribution.

kullmanp:
Thanks!

PR #149, opened 2019-03-01
BinaryUrlUtils fails to detect Mac/OSX

singhravik:
BinaryUrlUtils fails to detect Mac/OSX leaks into wrong section and returns Linux_64bit.

Logic fixed in `public static String getPlatformBinaryNameUrl() {}` method
arquillian-drone-saucelabs-extension/src/main/java/org/arquillian/drone/saucelabs/extension/utils/BinaryUrlUtils.java

IS_OS_MAC check now happens after IS_OS_UNIX==true.
For OSX. It is also a variation of Unix so both IS_OS_UNIX and IS_OS_MAC are 'true'. Handle OSX here.
**Fixes**: #


bartoszmajsak (approved):
Thank you for this fix.

PR #221, opened 2020-09-10
fix(download): case insensitive headers in HttpClient.Response

mkouba:
- header names should be case-insensitive
- also fallback to "now" when the Last-Modified header is absent

We're observing problems with downloading the latest Firefox driver due to the fact that a response from the GitHub API contains a header with name `last-modified` but drone expects `Last-Modified`. As a result a `NullPointerException` is thrown from a [ZonedDateTime.parse()](https://github.com/arquillian/arquillian-extension-drone/blob/master/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java#L145) invocation.

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java
@@ -141,8 +141,11 @@ public abstract class GitHubSource implements ExternalBinarySource {
 
     protected ZonedDateTime extractModificationDate(HttpClient.Response response) {
         final String modificationDate = response.getHeader(LAST_MODIFIED);
-        final DateTimeFormatter dateTimeFormatter = Rfc2126DateTimeFormatter.INSTANCE;
-        return ZonedDateTime.parse(modificationDate, dateTimeFormatter);
+        if (modificationDate != null) {
+            final DateTimeFormatter dateTimeFormatter = Rfc2126DateTimeFormatter.INSTANCE;
+            return ZonedDateTime.parse(modificationDate, dateTimeFormatter);
+        }
+        return ZonedDateTime.now();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


bartoszmajsak (approved):
Thanks @mkouba! I had to fix the travis build to let these builds run correctly. Once this passes and we merge it do you want me to cut the release?

manovotn:
@bartoszmajsak thanks for checking on this. This was an issue we bumped into with Weld test using Drone.

The release wouldn't solve the original issue we were seeing in CI env. Our CI was getting errors due to GH API limit exceed, so after a while I noticed [this part of documentation](http://arquillian.org/arquillian-extension-drone/#_authenticated_requests) which seems to be just the thing. However, that didn't work (neither locally, nor in CI env) and I will probably need to debug Drone to see why (once I figure which part of Drone to look into). I'd like to know for sure the error is not on Drone side (which is very well possible) before you cut a release. Alternatively, if you know a quick way to verify that, by all means go ahead ;-)

bartoszmajsak:
@manovotn To be honest with you I don't really know - I would need to dig into the code. But maybe @MatousJobanek still remembers? :)

MatousJobanek:
Hi, IIRC Drone used basic authentication, so maybe try to use different token or use password instead... Other than that, I have no clue :man_shrugging:

MatousJobanek:
Looking at the code quickly, it looks like it uses some [weird basic authorization header](https://github.com/arquillian/arquillian-extension-drone/blob/0de7f262378930f09f03fdc0d8ee082ac54569fc/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSource.java#L226-L231) - this is something I cannot find in GH documentation. So there are three options:
1. it's either a bug (which is weird because I'm sure that it was working before)
2. the GH auth has changed or they dropped something in last 3 years
3. I've been away from the code too long so I don't clearly understand the logic :grimacing:

manovotn:
I did try creating a token specifically for this which got me thinking it is probably not working. I am now a bit snowed under other stuff and due for some PTO on top of that (next week) but I'll try and keep this on my list to get back to it.
Thanks for those hint @MatousJobanek, now I at least know where to look :-)

Issue #259, opened 2021-04-25
GeckoDriverGitHubSource.getLatestRelease() when running on macOS x86_64 picks up geckodriver for wrong platform architecture


afedotov:
##### Issue Overview

When running Arquillian tests on macOS **x86_64** with Drone using Firefox browser, it downloads from GitHub geckodriver for wrong platform architecture - **aarch64** (new Macs with Apple M1 chip).

##### Expected Behaviour

Drone should download geckodriver for macOS according to current hardware architecture.

##### Steps To Reproduce

1. Clear Arquillian Drone local cache `rm -rf ~/.arquillian/drone/*`
2. Run simple `WebDriverTestCase#login` test using Firefox browser:
    `mvn clean verify -Dbrowser=firefox -Dtest=WebDriverTestCase#login -DfailIfNoTests=false`
3. Make sure that Drone downloads and trying to execute `geckodriver-v0.29.1-macos-aarch64.tar.gz` instead of `geckodriver-v0.29.1-macos.tar.gz`

##### Additional Information

<details>
 <summary>$ mvn clean verify -Dbrowser=firefox -Dtest=WebDriverTestCase#login -DfailIfNoTests=false</summary>
 <pre>
---8<--------------
[INFO] 
[INFO] --- maven-surefire-plugin:2.22.2:test (default-test) @ arquillian-drone-webdriver ---
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.jboss.arquillian.drone.webdriver.example.WebDriverTestCase
Apr 25, 2021 1:42:34 PM org.eclipse.jetty.util.log.Log initialized
INFO: Logging initialized @732ms to org.eclipse.jetty.util.log.JavaUtilLog
Apr 25, 2021 1:42:34 PM org.jboss.arquillian.container.jetty.embedded_9.JettyEmbeddedContainer start
INFO: Starting Jetty Embedded Server 9.4.7.v20170914 [id:888655833]
Apr 25, 2021 1:42:35 PM org.eclipse.jetty.server.Server doStart
INFO: jetty-9.4.7.v20170914
Apr 25, 2021 1:42:35 PM org.eclipse.jetty.server.AbstractConnector doStart
INFO: Started ServerConnector@7ea9e1e2{HTTP/1.1,[http/1.1]}{localhost:9090}
[...]

PR #261, opened 2021-05-08
fix: supports geckodriver binary for aarch64

afedotov:
**Fixes**: #259


bartoszmajsak (approved):
Thanks a lot for this valuable fix!

And apologies it took so long to merge.

PR #288, opened 2021-11-02
fix: handles Edge binary download from XML source

### LINKED ISSUE #274: getLatestVersion() from EdgeDriverBinaryHandler is not working
##### Issue Overview

When trying to run a test with Edge without using a binary file, Arquillian Drone fails to download the binary due to a change on Microsoft Edge web page

##### Expected Behaviour
Download the latest Edge binary or the one specified on the maven argument

##### Current Behaviour
IndexOutOfBounds error for an empty list due to not getting any version from the Microsoft Edge web page

##### Steps To Reproduce
1. Launch an Edge test without setting the binary file

bartoszmajsak:

Adjusts the logic of downloading MS Edge driver to be based on XML resource with all released drivers (and `LATEST_STABLE` release).

A general observation from working with the code:
- processing the whole XML tree is unnecessary for both Chrome and Edge drivers, are they provide access to the latest release versions through special links. With that, we could simply use XPath expressions to find matching nodes based on either a defined version or the latest released. 
- there are other driver handlers such as Internet Explorer and Selenium source which share the same abstract class. This makes the whole refactoring effort higher.

- new MS Edge driver downloading logic
- enhanced CI job which now includes a profile for this driver/browser
- minor renaming

Fixes #274


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * It is expected that this abstract method should return a regex that represents a key of an expected binary/file
+     * stored in the google storage. Key consist of <code>directory_name + / + file_name</code>. Keys are visible on
+     * the root of the google storage url (without index.html suffix)
+     *
+     * @param requiredVersion The required version set using method {@link XmlStorageSource#getReleaseForVersion},
+     *                        or otherwise null
+     * @param directory       The directory for the current binary the regex is being matched against
+     * @return A regex that represents a key of an expected binary/file stored in the google storage.
```
Aboullos, line 182: The description still refers to google storage

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
```diff
@@ -0,0 +1,217 @@
...
+     */
+    protected abstract String getExpectedKeyRegex(String requiredVersion, String directory);
```
Aboullos, line 184: The classes using this method only use the first parameter, the other one could be used in the future?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
```diff
@@ -74,16 +74,16 @@ protected String getExpectedKeyRegex(String requiredVersion, String directory) {
...
-                    fileName.append("64");
-                }
+            }
+            if (PlatformUtils.is32()) {
+                fileName.append("32");
+            } else {
+                fileName.append("64");
             }
             return fileName.append(".zip").toString();
         }
```
Aboullos, line 89: There is a similar one in `EdgeDriverBinary` but with a small difference in the order of the code and changing "chromedriver" with "edgedriver" could it be a joint method with the driver as another parameter in `XmlStorageSource`?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/EdgeDriverBinaryHandler.java
```diff
@@ -55,80 +56,64 @@ protected String getUrlToDownloadProperty() {
...
-                    break;
-                }
-            }
-
-            if (webDriverUrl.isEmpty()) {
-                log.warning("WebDriver with specified version was not found.");
-            }
+        protected String getExpectedKeyRegex(String requiredVersion, String directory) {
+            return Pattern.quote(requiredVersion + "/" + getFileNameRegexToDownload(requiredVersion));
+        }
```
Aboullos, line 75: Same as the other comment, it uses only the first parameter

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/EdgeDriverFactory.java
```diff
@@ -60,8 +61,9 @@ protected String getDriverReadableName() {
 
     public Capabilities getCapabilities(WebDriverConfiguration configuration, boolean performValidations) {
         DesiredCapabilities capabilities = new DesiredCapabilities(configuration.getCapabilities());
-        capabilities.setPlatform(DesiredCapabilities.edge().getPlatform());
+        capabilities.setPlatform(Platform.ANY);
         capabilities.setBrowserName(DesiredCapabilities.edge().getBrowserName());
+//        capabilities.setCapability("useChromium", "true");
```
Aboullos, line 66: Commented line

Aboullos (commented):
Looks a lot better than my draft! It is working right for Windows, haven't check yet for Linux

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
```diff
@@ -0,0 +1,217 @@
...
+
+    protected String nodeName;
+
+    protected String fileName;
+
+    private String storageUrl;
+
+    private String urlToLatestRelease;
+
+    private List<DriverEntry> driverEntries;
```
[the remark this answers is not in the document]
bartoszmajsak: That's true, I just changed the type here from `ArrayList` though :)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
bartoszmajsak: I'm not sure, I think this should be kept per driver binary basis. In my eyes it's going a bit too far with "reuse", but I will think it over. Thanks for spotting that!

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/EdgeDriverFactory.java
bartoszmajsak: Right :) That's when I was testing Edge for Linux. You could use suggestions here to remove:

```suggestion
```

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
bartoszmajsak:
```suggestion
     * It is expected that this abstract method should return a regex that represents a key of an expected binary/file
     * stored in the xml storage. Commonly, key consists of <code>directory_name + / + file_name</code>. 
     *
     * @param requiredVersion The required version set using method {@link XmlStorageSource#getReleaseForVersion},
     *                        or otherwise null
     * @param directory       The directory for the current binary the regex is being matched against
     * @return A regex that represents a key of an expected binary/file.
```

Is it better?

bartoszmajsak:
> Looks a lot better than my draft! It is working right for Windows, haven't check yet for Linux

Thanks for the review and verification against your test suite. I will go over your suggestions tomorrow.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
Aboullos: Okay, or maybe if it is not reused they could have the same code, the condition are in a different order (althought the result is the same, but to have the same approach on both)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
bartoszmajsak: This is why I was not particularly happy about the current structure here. It's a bit mixed bag of responsibilities and structure pushed up in the hierarchy. It leads to this case which caught your attention it's used in 2 out of 4 implementations - `SeleniumServerStorage` and  `InternetExplorerBinaryHandler` both rely on `directory`. I will keep it as is for now, but I feel this solution can be simplified if we decide to redo it with XPath approach.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
bartoszmajsak: Difference here is however that in case of Edge Driver there is no such thing as `mac32` thus for Mac it's always rolling back to `mac64.zip` suffix. This part (apart from filename prefix) makes those implementations different. I think we should just leave it as is.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/EdgeDriverBinaryHandler.java
bartoszmajsak: See https://github.com/arquillian/arquillian-extension-drone/pull/288#discussion_r741930672

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
```diff
@@ -0,0 +1,217 @@
...
+                driverEntry.setLocation(getLocation(key, item));
+                results.add(driverEntry);
+            }
+        }
+        return results;
+    }
+
+    protected String getLocation(String key, Element item) {
+        return key.substring(0, key.indexOf("/"));
+    }
```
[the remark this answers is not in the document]
bartoszmajsak: Addressed in ad1ae09.

Aboullos (approved):
LGTM!

Issue #296, opened 2021-11-23
LATEST_STABLE Edge driver might not be available for Linux


bartoszmajsak:
##### Issue Overview

Every now and then when there's new MS Edge Driver released we are seeing build errors, because Linux driver is not (yet?) released.

At this moment calling https://msedgewebdriverstorage.blob.core.windows.net/edgewebdriver/LATEST_STABLE returns `96.0.1054.29` which is not on the download list we rely on https://msedgedriver.azureedge.net/

*Available driver for LATEST_STABLE=96.0.1054.34*

```
   <Blob>
      <Name>96.0.1054.29/edgedriver_arm64.zip</Name>
      <Url>https://msedgewebdriverstorage.blob.core.windows.net/edgewebdriver/96.0.1054.29/edgedriver_arm64.zip</Url>
      <Properties>
         <Last-Modified>Sat, 20 Nov 2021 01:52:05 GMT</Last-Modified>
         <Etag>0x8D9ABC85A9743EB</Etag>
         <Content-Length>6835653</Content-Length>
         <Content-Type>application/octet-stream</Content-Type>
         <Content-Encoding />
         <Content-Language />
         <Content-MD5>qJx+KvI7CWrXtL5TO9eSvQ==</Content-MD5>
         <Cache-Control />
         <BlobType>BlockBlob</BlobType>
         <LeaseStatus>unlocked</LeaseStatus>
      </Properties>
   </Blob>
   <Blob>
      <Name>96.0.1054.29/edgedriver_win32.zip</Name>
      <Url>https://msedgewebdriverstorage.blob.core.windows.net/edgewebdriver/96.0.1054.29/edgedriver_win32.zip</Url>
      <Properties>
         <Last-Modified>Sat, 20 Nov 2021 01:52:05 GMT</Last-Modified>
         <Etag>0x8D9ABC85AB529AB</Etag>
         <Content-Length>7004938</Content-Length>
         <Content-Type>application/octet-stream</Content-Type>
         <Content-Encoding />
         <Content-Language />
         <Content-MD5>mmc0ObYDuO+MPy7gk4ujAQ==</Content-MD5>
         <Cache-Control />
         <BlobType>BlockBlob</BlobType>
         <LeaseStatus>unlocked</LeaseStatus>
      </Properties>
   </Blob>
   <Blob>
      <Name>96.0.1054.29/edgedriver_win64.zip</Name>
[...]

PR #297, opened 2021-11-25
fix(edge): downloads  OS release when latest stable not available

bartoszmajsak:

Introduces fallback mechanism if MS Edge driver binary in the version published as `LATEST_STABLE` is not available for a given OS. It downloads the one published as `LATEST_RELEASE_X_OS` where `X` is a version, e.g. `96` (aligned with the `STABLE` stream) and `OS` is `WINDOWS`, `MACOS` or `LINUX`.

- enhanced download logic for edge driver
- additional tests to cover this path

Fixes: #296


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
```diff
@@ -76,14 +77,20 @@ public ExternalBinary getLatestRelease() throws Exception {
...
-            latestVersion = StringUtils.trimMultiline(httpClient.get(urlToLatestRelease, charset).getPayload());
+            latestVersion = getVersion(urlToLatestRelease, charset);
         }
         return getReleaseForVersion(latestVersion);
     }
 
+    protected String getVersion(String urlToLatestRelease, String charset) throws IOException {
+        return StringUtils.trimMultiline(httpClient.get(urlToLatestRelease, charset).getPayload());
+    }
+
```
bartoszmajsak, line 88:
```suggestion
    protected String getVersion(String versionUrl, String charset) throws IOException {
        return StringUtils.trimMultiline(httpClient.get(versionUrl, charset).getPayload());
    }

```

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
```diff
@@ -76,14 +77,20 @@ public ExternalBinary getLatestRelease() throws Exception {
...
     private List<DriverEntry> retrieveAllDriversEntries() throws Exception {
         final List<DriverEntry> results = new ArrayList<>();
-        final DocumentBuilder db = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+        final DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+        dbf.setIgnoringElementContentWhitespace(true);
+        final DocumentBuilder db = dbf.newDocumentBuilder();
```
bartoszmajsak, line 93:
```suggestion
        final DocumentBuilder db = DocumentBuilderFactory.newInstance().newDocumentBuilder();
```

PR #301, opened 2021-12-08
fix: falls back to 32-bit chrome driver Windows

### LINKED ISSUE #300: Chrome driver 64-bit might not be available for Windows
##### Issue Overview

It's not possible to fetch the Chrome driver for Windows machines due to the fact, there's only provided 32-bit version of the driver and Drone is seeking for the 64-bit version. However, the 32-bit version should be compatible to 64-bit browsers and machines [1].

[1] https://bugs.chromium.org/p/chromedriver/issues/detail?id=1797#c1

#### Solution
Solution for that would be that if the 64-bit WebDriver for Windows is not found, it'd try to fetch the 32-bit version of it.

#### Provided Chrome Drivers

```
 <Contents>
<Key>96.0.4664.18/chromedriver_linux64.zip</Key>
<Generation>1635142532134035</Generation>
<MetaGeneration>1</MetaGeneration>
<LastModified>2021-10-25T06:15:32.167Z</LastModified>
<ETag>"e8b409daf617e69cbf9a325606d65375"</ETag>
<Size>9981693</Size>
</Contents>
<Contents>
<Key>96.0.4664.18/chromedriver_mac64.zip</Key>
<Generation>1635142534339256</Generation>
<MetaGeneration>1</MetaGeneration>
<LastModified>2021-10-25T06:15:34.371Z</LastModified>
<ETag>"6ce0a3b645e8bc4d0baf4d21fd0fa902"</ETag>
<Size>8232190</Size>
</Contents>
<Contents>
<Key>96.0.4664.18/chromedriver_mac64_m1.zip</Key>
<Generation>1635142536513349</Generation>
<MetaGeneration>1</MetaGeneration>
<LastModified>2021-10-25T06:15:36.546Z</LastModified>
<ETag>"87b3bf1f8bac7df115930330b3eaf6a9"</ETag>
<Size>7787623</Size>
</Contents>
<Contents>
<Key>96.0.4664.18/chromedriver_win32.zip</Key>
<Generation>1635142538618432</Generation>
<MetaGeneration>1</MetaGeneration>
<LastModified>2021-10-25T06:15:38.651Z</LastModified>
<ETag>"5201972f802fcb345fc93cb4ee117683"</ETag>
<Size>6073766</Size>
</Contents>
```

#### Error
```
Caused by: org.jboss.arquillian.drone.webdriver.binary.downloading.source.MissingBinaryException: There wasn't found any binary with the key matching regex \Q96.0.4664.45/chromedriver_win64.zip\E in the storage: https://chromedriver.storage.googleapis.com/
	at org.jboss.arquillian.drone.webdriver.binary.downloading.source.XmlStorageSource.getReleaseForVersion(XmlStorageSource.java:157)
	at org.jboss.arquillian.drone.webdriver.binary.downloading.source.XmlStorageSource.getLatestRelease(XmlStorageSource.java:82)
	at org.jboss.arquillian.drone.webdriver.binary.downloading.source.XmlStorageSource.getLatestRelease(XmlStorageSource.java:75)
	at org.jboss.arquillian.drone.webdriver.binary.handler.AbstractBinaryHandler.downloadAndPrepare(AbstractBinaryHandler.java:175)
	at org.jboss.arquillian.drone.webdriver.binary.handler.AbstractBinaryHandler.checkAndSetBinary(AbstractBinaryHandler.java:61)
	... 12 more
```

mabartos:
When 64-bit is not available.

Closes #300


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/ExternalBinarySource.java
```diff
@@ -40,4 +59,17 @@
      * @return A regex that represents a file name of an asset to be downloaded.
      */
     String getFileNameRegexToDownload(String version);
+
+    /**
+     * This method should returns a regex that represents a file name of the release asset that should be downloaded.
+     * If this method returns null, the AbstractBinaryHandler won't be looking for any file in the cache directory
+     * and will jump to downloading phase.
```
bartoszmajsak, line 66: I'm not convinced we should document any behaviour of anything implementing that interface. I would simply remove it.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
```diff
@@ -60,18 +67,52 @@ public String getSystemBinaryProperty() {
...
+            return latestRelease;
         }
 
         @Override
         public String getFileNameRegexToDownload(String version) {
+            return getFileNameRegexToDownload(version, null);
+        }
+
+        @Override
+        public String getFileNameRegexToDownload(String version, String systemVersion) {
```
bartoszmajsak, line 115: Another reason to have something more meaningful.

I think enum `Platform` with `_32`, `_64` and `AUTO_DETECT` should work better.

Something like:

```java
public enum Platform {
    _32("32"),
    _64("64"),
    AUTO_DETECT("auto-detect"){
        @Override
        public String getPlatformCode() {
            return PlatformUtils.is32() ? "32" : "64";
        }
    };
    

    private String code;

    Platform(String code) {
        this.code = code;
    }
    
    public String getPlatformCode() {
        return this.code;
    }
}
```

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/ExternalBinarySource.java
```diff
@@ -32,6 +32,25 @@
...
     ExternalBinary getReleaseForVersion(String version) throws Exception;
 
+    /**
+     * Retrieves information about a binary release with the desired version and returns an instance of
+     * {@link ExternalBinary} that represents the found release.
+     *
+     * @param version
+     *     A version of a binary release that should be retrieved.
+     *
+     * @param systemVersion 32 or 64
```
bartoszmajsak, line 42: I would call it `platform`.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/XmlStorageSource.java
```diff
@@ -141,13 +140,17 @@ protected String getContentOfFirstElement(Element element, String tagName) {
 
     @Override
     public ExternalBinary getReleaseForVersion(String requiredVersion) throws Exception {
+        return getReleaseForVersion(requiredVersion, null);
```
bartoszmajsak, line 143: `null` no likey :) See comment down below.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/StubbedChromeStorageTestCase.java
```diff
@@ -0,0 +1,69 @@
...
+        // when
+        final ExternalBinary latestRelease = chromeDrivers.getLatestRelease();
+
+        // then
+        assertThat(latestRelease.getVersion(), is((ALL_OS_RELEASE)));
+    }
+
+    @Test
+    public void should_find_latest_stable_release_with_32_bit_version() throws Exception {
+        System.setProperty("os.name", "win");
```
bartoszmajsak, line 60: This way it is set to Windows for all the tests coming later on the same machine... That is why tests for Edge are failing. You should clean up this value after the test execution. See for example https://stefanbirkner.github.io/system-rules/index.html#RestoreSystemProperties

I am also a bit concerned because it could lead to test failures if there are two jobs running at the same time, as it will alternate system-wide value. If we really want to test it on "every platform" we should find a better way than messing up system properties. Humm...

humm

mabartos: Thanks for the example; it's very useful. I kept the settings of the System property for now and we can discuss the best solution for that lately.

mabartos:
@bartoszmajsak Thanks for your review! I replaced the parameter `systemVersion` by enum `Architecture`. The enum was already included in the code base and I'd say it's better than name `Platform` for that. Platform might represent whole platform including OS and Architecture itself; as is described in `PlatformUtils` class. WDYT? However, if you'd like to have it named as `Platform`, I can change it :))

bartoszmajsak:
Sure, go ahead :)

bartoszmajsak (approved):

PR #308, opened 2022-01-05
fix: falls back to Windows 32bit chrome driver release

jomarko:
For more details see:
- https://github.com/arquillian/arquillian-extension-drone/issues/300

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
@@ -10,6 +10,7 @@ import org.jboss.arquillian.drone.webdriver.utils.HttpClient;
 import org.jboss.arquillian.drone.webdriver.utils.PlatformUtils;
 import org.openqa.selenium.remote.DesiredCapabilities;
 
+import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
@@ -85,6 +86,25 @@ public class ChromeDriverBinaryHandler extends AbstractBinaryHandler {
         @Override
         protected String getExpectedKeyRegex(String requiredVersion, String directory) {
             return getExpectedKeyRegex(requiredVersion, directory, Architecture.AUTO_DETECT);
+        }
+
+        @Override
+        public ExternalBinary getReleaseForVersion(String requiredVersion, Architecture architecture) throws Exception {
+            ExternalBinary release;
+            try {
+                release = super.getReleaseForVersion(requiredVersion, architecture);
+            } catch (MissingBinaryException mbe) {
+                if (PlatformUtils.isWindows() && Objects.equals(architecture.getValue(), Architecture.BIT64.getValue())) {
+                    log.log(Level.WARNING, "Failed downloading 64-bit version of Chrome Driver. Reason: ", mbe);
+                    log.log(Level.WARNING, "This special case was reported as https://github.com/arquillian/arquillian-extension-drone/issues/300");
+                    log.log(Level.WARNING, "Downloading 32-bit version of Chrome Driver instead. ({0})", requiredVersion);
+
+                    release = getReleaseForVersion(requiredVersion, Architecture.BIT32);
+                } else {
+                    throw mbe;
+                }
+            }
+            return release;
         }
 
         @Override
--- a/drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/StubbedChromeStorageTestCase.java
+++ b/drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/StubbedChromeStorageTestCase.java
@@ -74,7 +74,7 @@ public class StubbedChromeStorageTestCase {
     }
 
     @Test
-    public void should_always_fallback_to_32_bit_version() throws Exception {
+    public void on_any_windows_should_download_32_bit_chromedriver() throws Exception {
         System.setProperty("os.name", "win");
 
         assertThat(PlatformUtils.isWindows(), is(true));
@@ -82,14 +82,28 @@ public class StubbedChromeStorageTestCase {
         final ExternalBinary releaseWithArchitectureDetected = chromeDrivers.getReleaseForVersion(ALL_OS_RELEASE);
         assertThat(releaseWithArchitectureDetected.getVersion(), is(ALL_OS_RELEASE));
         assertThat(releaseWithArchitectureDetected.getUrl(), CoreMatchers.endsWith("win32.zip"));
+    }
 
-        final ExternalBinary release32 = chromeDrivers.getReleaseForVersion(ALL_OS_RELEASE, Architecture.BIT32);
-        assertThat(release32.getVersion(), is(ALL_OS_RELEASE));
-        assertThat(release32.getUrl(), CoreMatchers.endsWith("win32.zip"));
+    @Test
+    public void on_windowns32b_should_download_32_bit_chromedriver() throws Exception {
+        System.setProperty("os.name", "win");
 
-        final ExternalBinary release64 = chromeDrivers.getReleaseForVersion(ALL_OS_RELEASE, Architecture.BIT64);
-        assertThat(release64.getVersion(), is(ALL_OS_RELEASE));
-        assertThat(release64.getUrl(), CoreMatchers.endsWith("win32.zip"));
+        assertThat(PlatformUtils.isWindows(), is(true));
+
+        final ExternalBinary release = chromeDrivers.getReleaseForVersion(ALL_OS_RELEASE, Architecture.BIT32);
+        assertThat(release.getVersion(), is(ALL_OS_RELEASE));
+        assertThat(release.getUrl(), CoreMatchers.endsWith("win32.zip"));
+    }
+
+    @Test
+    public void on_windowns64b_should_download_32_bit_chromedriver() throws Exception {
+        System.setProperty("os.name", "win");
+
+        assertThat(PlatformUtils.isWindows(), is(true));
+
+        final ExternalBinary release = chromeDrivers.getReleaseForVersion(ALL_OS_RELEASE, Architecture.BIT64);
+        assertThat(release.getVersion(), is(ALL_OS_RELEASE));
+        assertThat(release.getUrl(), CoreMatchers.endsWith("win32.zip"));
     }
 
     @Test(expected = MissingBinaryException.class)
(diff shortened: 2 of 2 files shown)


jomarko:
@bartoszmajsak may I ask you for a review?

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
```diff
@@ -87,6 +88,25 @@ protected String getExpectedKeyRegex(String requiredVersion, String directory) {
             return getExpectedKeyRegex(requiredVersion, directory, Architecture.AUTO_DETECT);
         }
 
+        @Override
+        public ExternalBinary getReleaseForVersion(String requiredVersion, Architecture architecture) throws Exception {
+            ExternalBinary release;
+            try {
+                release = super.getReleaseForVersion(requiredVersion, architecture);
+            } catch (MissingBinaryException mbe) {
+                if (PlatformUtils.isWindows() && Objects.equals(architecture.getValue(), Architecture.BIT64.getValue())) {
```
bartoszmajsak, line 97: Can't we say `Architecture.BIT64 == architecture` for enums?

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/StubbedChromeStorageTestCase.java
```diff
@@ -70,4 +72,32 @@ public void should_find_latest_stable_release_with_32_bit_version() throws Excep
...
+        assertThat(releaseWithArchitectureDetected.getVersion(), is(ALL_OS_RELEASE));
+        assertThat(releaseWithArchitectureDetected.getUrl(), CoreMatchers.endsWith("win32.zip"));
+
+        final ExternalBinary release32 = chromeDrivers.getReleaseForVersion(ALL_OS_RELEASE, Architecture.BIT32);
+        assertThat(release32.getVersion(), is(ALL_OS_RELEASE));
+        assertThat(release32.getUrl(), CoreMatchers.endsWith("win32.zip"));
+
+        final ExternalBinary release64 = chromeDrivers.getReleaseForVersion(ALL_OS_RELEASE, Architecture.BIT64);
+        assertThat(release64.getVersion(), is(ALL_OS_RELEASE));
+        assertThat(release64.getUrl(), CoreMatchers.endsWith("win32.zip"));
```
bartoszmajsak, line 92: Why not split it into 3 separate test cases?

bartoszmajsak (changes requested):
Thanks - just two small suggestions from my side.

And one big question - why do we prefer to always fall back if there's 64bit available?

bartoszmajsak:
Interesting to see the failure for `Edge` build:

```
Caused by: org.jboss.arquillian.drone.webdriver.binary.downloading.source.MissingBinaryException: There wasn't found any binary with the key matching regex \Q96.0.1054.62/edgedriver_linux64.zip\E in the storage: https://msedgedriver.azureedge.net/
	at org.jboss.arquillian.drone.webdriver.binary.downloading.source.XmlStorageSource.getReleaseForVersion(XmlStorageSource.java:162)
	at org.jboss.arquillian.drone.webdriver.binary.downloading.source.XmlStorageSource.getReleaseForVersion(XmlStorageSource.java:145)
	at org.jboss.arquillian.drone.webdriver.binary.handler.EdgeDriverBinaryHandler$EdgeStorageSources.getLatestRelease(EdgeDriverBinaryHandler.java:101)
	at org.jboss.arquillian.drone.webdriver.binary.handler.EdgeDriverBinaryHandler$EdgeStorageSources.getLatestRelease(EdgeDriverBinaryHandler.java:89)
	at org.jboss.arquillian.drone.webdriver.binary.handler.AbstractBinaryHandler.downloadAndPrepare(AbstractBinaryHandler.java:175)
	at org.jboss.arquillian.drone.webdriver.binary.handler.AbstractBinaryHandler.checkAndSetBinary(AbstractBinaryHandler.java:61)
	... 13 more
```

as it's clearly available:

```xml
<Blob>
<Name>96.0.1054.62/edgedriver_linux64.zip</Name>
<Url>https://msedgewebdriverstorage.blob.core.windows.net/edgewebdriver/96.0.1054.62/edgedriver_linux64.zip</Url>
<Properties>
<Last-Modified>Fri, 17 Dec 2021 17:49:49 GMT</Last-Modified>
<Etag>0x8D9C1859F731EF2</Etag>
<Content-Length>11941648</Content-Length>
<Content-Type>application/octet-stream</Content-Type>
<Content-Encoding/>
<Content-Language/>
<Content-MD5>TQ3ildCw4nwjLWYdnUREZw==</Content-MD5>
<Cache-Control/>
<BlobType>BlockBlob</BlobType>
<LeaseStatus>unlocked</LeaseStatus>
</Properties>
</Blob>
```

Re-running the job.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
jomarko: Yes, good point. That was my first assumption too. Unfortunately we have a third constant `Architecture.AUTO_DETECT` that is dynamically evaluated and may point as to `32` so to `64` bit version.

This third constant is used for example in `XmlStorageSource`
```
    @Override
    public ExternalBinary getReleaseForVersion(String requiredVersion) throws Exception {
        return getReleaseForVersion(requiredVersion, Architecture.AUTO_DETECT);
    }
```

In other words, `Architecture.BIT64 == architecture` is evalueated to false on `64` machine, if `AUTO_DETECT` was used.

bartoszmajsak:
Ok I'm lost :D 

https://app.circleci.com/pipelines/github/arquillian/arquillian-extension-drone/120/workflows/d3aad4a3-3c9b-493c-ac31-f28e06f8938b/jobs/404?invite=true#step-109-1871

bartoszmajsak:
And it works on my machine....

```
Jan 05, 2022 5:13:17 PM org.jboss.arquillian.drone.webdriver.binary.BinaryFilesUtils extract
INFO: Extracting zip file: /home/bartek/.arquillian/drone/edge/96.0.1054.62/edgedriver_linux64.zip to /home/bartek/code/redhat/arquillian/arquillian-extension-drone/drone-webdriver/target/drone/4d0de295d0b0e27c232d661d9d444467
```

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
bartoszmajsak: Ah right!

bartoszmajsak:
After a little bit of investigation, I think I know what is going on. Here's what I did:

* ran affected test locally -> `success`
* started failing job with SSH and login to the node
* ran the same test again there -> `failure`
* curl https://msedgedriver.azureedge.net/ to see what is there -> the version was included
* re-ran the same test again from the node -> `success`

This leads me to the conclusion that it's likely some cache in-between (circle-ci perhaps?) which caches the result. Quick fix would be to set HTTP headers to control the cache:

```
"Cache-Control": "no-cache" // HTTP 1.1
"Pragma": "no-cache" // HTTP 1.0
```

But the better approach would be to rework the `XmlStorage` logic. We get the latest versions from some arbitrary URLs and we also know how to construct the URLs to get the correct version of the driver. We don't have to parse that XML file for a given version* to validate if it's there.

WDYT?

\* In fact we don't have to have this blob at all :)

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
bartoszmajsak: @jomarko I just realized my comment could have been perceived as rude which was not my intention at all. Apologies.

jomarko: @bartoszmajsak not at all. No need to apologize. Do not worry!

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/StubbedChromeStorageTestCase.java
jomarko: yes, I will split it

jomarko: see latest commit

jomarko:
Regarding https://github.com/arquillian/arquillian-extension-drone/pull/308#issuecomment-1005898875 yes, I am open to help with this, but I believe it should be done in separate PR.

jomarko:
Regarding _And one big question - why do we prefer to always fall back if there's 64bit available?_ I believe it is just my not appropriate wording in PR? Reading my code, I believe 64b chrome driver will be downloaded if available. Should I add test for it?

bartoszmajsak:
> Regarding [#308 (comment)](https://github.com/arquillian/arquillian-extension-drone/pull/308#issuecomment-1005898875) yes, I am open to help with this, but I believe it should be done in separate PR.

That would be awesome, and of course, I didn't assume it will go as this PR. It's unrelated.

bartoszmajsak:
> Regarding _And one big question - why do we prefer to always fall back if there's 64bit available?_ I believe it is just my not appropriate wording in PR? Reading my code, I believe 64b chrome driver will be downloaded if available. Should I add test for it?

I got that impression from `on_any_windows_should_download_32_bit_chromedriver`. From the code itself, it is a fallback mechanism clearly. I think the test itself is not revealing the nitty-gritty detail that we only have `96.0.4664.18/chromedriver_win32.zip` in the test XML :) I wonder if renaming to `ALL_OS_BUT_WIN64_RELEASE` would be sufficient clarification?

jomarko:
@bartoszmajsak yes I tried to improve the naming in the test, see latest commit please.

bartoszmajsak (approved):
Excellent work! Thank you very much @jomarko!

jomarko:
it seems Ci is green, can we merge this PR?

bartoszmajsak:
> it seems Ci is green, can we merge this PR?

we are hitting the edge issue again after the rebase. on it :)

bartoszmajsak:
@jomarko  Both `LATEST_STABLE` and `LATEST_RELEASE_97_LINUX` point to `97.0.1072.55` which is nowhere to be found on the list https://msedgedriver.azureedge.net/

It is downloadable however https://msedgewebdriverstorage.blob.core.windows.net/edgewebdriver/97.0.1072.55/edgedriver_linux64.zip :)

I think we should re-work (and simplify) the way we handle it. We don't need that XML file at all

bartoszmajsak:
I'll merge it now and hopefully, we will fix issue with edge soon :)

Issue #309, opened 2022-01-06
Improve XmlStorage mechanism


jomarko:
##### Issue Overview

As discussed in PR https://github.com/arquillian/arquillian-extension-drone/pull/308#issuecomment-1005898875, we have an opportunity to improve internal mechanism of driver version detection/validation.

jomarko:
@bartoszmajsak I created this issue, can I somehow assign it to me?

bartoszmajsak:
@jomarko done :)

bartoszmajsak:
I think we should simplify the download logic - see my recent comment on your other PR https://github.com/arquillian/arquillian-extension-drone/pull/308#issuecomment-1007297112

bartoszmajsak:
One more thing btw - if we manage to have similar logic for both Edge and Chrome, meaning we know how to construct URLs to fetch drivers and have URLs for version/os which gives us the version number, there's no reason to call it `XmlStorage`  I believe.

bartoszmajsak:
Fixed in #310

PR #310, opened 2022-01-07
feat: reworks Chrome/Edge driver download to avoid XML processing

jomarko:
Changes in this PR tries to download browser webdrivers more straightforwardly. We do not parse xml with all versions available anymore, as we can construct the url for downloading particular webdriver without it.

If the webdriver is available, response has `200` status code.
If the webdriver is not available, response has `404` status code.

`ChromeDriverBinaryHandler` and`EdgeDriverBinaryHandler` do not more extend `XmlStorageSource` to avoid xml file parsing. The binary url can be derived without xml parsing for these browsers.

**Fixes**:
- #309


drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/UrlStorageSource.java
```diff
@@ -0,0 +1,90 @@
...
+    public ExternalBinary getReleaseForVersion(String requiredVersion) throws Exception {
+        return getReleaseForVersion(requiredVersion, Architecture.AUTO_DETECT);
+    }
+
+    @Override
+    public ExternalBinary getReleaseForVersion(String requiredVersion, Architecture architecture) throws Exception {
+        final String externalBinaryUrl = storageUrl + getFileNameRegexToDownload(requiredVersion, architecture);
+
+        final HttpClient.Response response = httpClient.get(externalBinaryUrl);
+        if (response.getStatusCode() == 404) {
```
bartoszmajsak, line 83: What if the server responds with e.g. `503` or even `403`? I would rather verify if we succeeded (so we got `200`) and return the binary but fail with all the other cases.

bartoszmajsak (commented):
Excellent! Thank you very much. I have one question before we merge it.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/UrlStorageSource.java
jomarko: thank you, I will update the PR

bartoszmajsak (approved):
Great work @jomarko!

PR #311, opened 2022-01-12
fix: fails when previously downloaded file doesn't exist

krichter722:
Throw `IllegalStateException` in case the downloaded file passed to `AbstractBinaryHandler.prepare` doesn't exist in order to avoid less intuitive error message during extraction.

Non-intuitive error message in case a downloaded binary vanishes after download

- Throw IllegalStateException if downloaded file vanished


bartoszmajsak (approved):
Thank you for your contribution!

PR #327, opened 2022-04-03
feat(chrome): downloads driver version for Apple Silicon

lainsr:
Use the M1 build of the Goggle Chrome WebDriver on Apple Silicon

There is an extra build of Google Chrome WebDriver for Apple Silicon. The name of the binary has an additional _m1 at the end. The x64 driver essentially works but starts Chrome in Rosetta. The M1 binary speeds the tests quit significantly.


bartoszmajsak (approved):
Awesome, thank you very much. Let me see why Firefox (?) build is failing and I'll merge.

PR #373, opened 2023-01-11
fix(chromedriver): ensures to extract binary

stringbean:
#### Short description of what this resolves:

ChromeDriver 109.0.5414.25 changed the contents of the driver archive to include a license file (`LICENSE.chromedriver`). When `ChromeDriverBinaryHandler` prepares the driver it attempts to make this license file executable which causes an exception to be raised:

```text
INFO: Extracting zip file: /home/jenkins/.arquillian/drone/chrome/109.0.5414.74/chromedriver_linux64.zip to /home/jenkins/workspace/example/target/drone/245747a6d5f2b7da02c465941f43e0e8
Jan 11, 2023 11:39:47 AM org.jboss.arquillian.drone.webdriver.binary.handler.AbstractBinaryHandler markAsExecutable
INFO: marking binary file: /home/jenkins/workspace/example/target/drone/245747a6d5f2b7da02c465941f43e0e8/LICENSE.chromedriver as executable
/home/jenkins/workspace/example/target/drone/245747a6d5f2b7da02c465941f43e0e8/LICENSE.chromedriver: line 1: //: Is a directory
/home/jenkins/workspace/example/target/drone/245747a6d5f2b7da02c465941f43e0e8/LICENSE.chromedriver: line 2: //: Is a directory
/home/jenkins/workspace/example/target/drone/245747a6d5f2b7da02c465941f43e0e8/LICENSE.chromedriver: line 3: //: Is a directory
...
```

The current behaviour grabs the first file in the extracted archive - on our CI system this happens to be the license file and not the driver.

#### Changes proposed in this pull request:

This PR changes the logic to find the ChromeDriver binary by filename instead of grabbing the first file.

PR #374, opened 2023-01-11
fix: Use ARM version of ChromeDriver on Apple Silicon

stringbean:

On Apple Silicon Macs the Intel version of ChromeDriver (`chromedriver_mac64.zip`) is used instead of the ARM version (`chromedriver_mac_arm64.zip`).

This PR adds a check for `isMacAppleSilicon` and inserts `_arm` into the filename if matched.

### DIFF
--- a/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
+++ b/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
@@ -137,29 +137,170 @@ public class ChromeDriverBinaryHandler extends AbstractBinaryHandler {
             return getFileNameRegexToDownload(version, Architecture.AUTO_DETECT);
         }
 
+        private static final ChromeVersion LAST_LINUX_32BIT = new ChromeVersion(2, 33, 0, 0);
+        private static final ChromeVersion LAST_MAC_32BIT = new ChromeVersion(2, 22, 0, 0);
+        private static final ChromeVersion FIRST_MAC_64BIT = new ChromeVersion(2, 23, 0, 0);
+        private static final ChromeVersion FIRST_MAC_M1 = new ChromeVersion(88, 0, 0, 0);
+        private static final ChromeVersion LAST_MAC_M1 = new ChromeVersion(106, 0, 5249, 0);
+
         @Override
         public String getFileNameRegexToDownload(String version, Architecture architecture) {
+            // known formats:
+            // chromedriver_linux32.zip    v <= 2.33
+            // chromedriver_linux64.zip
+            // chromedriver_win32.zip
+            // chromedriver_mac32.zip      v <= 2.22
+            // chromedriver_mac64.zip      v >= 2.23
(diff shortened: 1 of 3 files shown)


bartoszmajsak:
@stringbean could you rebase this PR so we get rid of conflicts and only have your changes from [40edd5f](https://github.com/arquillian/arquillian-extension-drone/pull/374/commits/40edd5fc3020b71e3a9e0b70821003eb0be5e3cf)?

stringbean:
Apologies I developed the fix against the `v2-support` branch and raised the PR against `master`.

drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java
```diff
@@ -127,6 +127,9 @@ public String getFileNameRegexToDownload(String version, Architecture architectu
             fileName.append("chromedriver_");
             if (PlatformUtils.isMac()) {
                 fileName.append("mac");
+                if (PlatformUtils.isMacAppleSilicon()) {
+                    fileName.append("_arm");
+                }
```
bartoszmajsak, line 132: On line [`141`](https://github.com/stringbean/arquillian-extension-drone/blob/c4105771715c151a8f70040e8677e16f81cd0f2c/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/binary/handler/ChromeDriverBinaryHandler.java#L141-L143) there is also the following check:

```
if (PlatformUtils.isMacAppleSilicon()) {
	fileName.append("_m1");
}
```
Wouldn't it result in a wrong filename - with `_m1` added at the end?

I think the naming schema changed between the versions and if the version will be pinned to e.g. [`102.0.5005.61`](https://chromedriver.storage.googleapis.com/index.html?path=102.0.5005.61/) it will attempt to download the non-existing file.

The logic needs to handle such a case. It seems the last version using `_m1` suffix is [`106.0.5249.21`](https://chromedriver.storage.googleapis.com/index.html?path=106.0.5249.21/). We might want to consider a fallback with the suffix instead of assuming that from version X the new naming is used. Just an idea.

bartoszmajsak (changes requested):
Thanks for investigating the problem. I'm afraid we need a bit more logic in place due to name changes.

bartoszmajsak:
@stringbean I will cut the release with your #373 fix and when this one gets finalized we will cut another one.

lprimak:
@bartoszmajsak @stringbean Can someone else verify that 14 works correctly? I am getting the same incorrect results as I was getting with 13.  Thank you

bartoszmajsak:
> Can someone else verify that 14 works correctly

@lprimak What do you mean by `14`?

lprimak:
@bartoszmajsak Arquillian 14 -> shorthand for 1.7.0.Alpha14
 :)

stringbean:
@bartoszmajsak I've extended the driver selection logic to handle all the (current) scenarios and added some unit tests to verify the results. Hopefully the Chrome team won't change the naming scheme for a while.

lprimak:
@bartoszmajsak bump please :)

lprimak:
@bartoszmajsak another bump please :)

PR #403, opened 2023-08-08
Remove support for PhantomJS

### LINKED ISSUE #402: Remove support for PhantomJS
##### Issue Overview

PhantomJS development is suspended and I believe, it doesn't worth maintaining it for Arquillian Drone.

https://github.com/ariya/phantomjs#phantomjs---scriptable-headless-webkit

mabartos:
Closes #402

Removal of all workarounds, components, and integrations related to the PhantomJS.

@bartoszmajsak Could you please check it?


mabartos:
@bartoszmajsak Can you check it, please?

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSourceLatestReleaseFromTagsTestCase.java
bartoszmajsak, line 1: I don't think this test should be removed simply because it's using PhantomJS as a repository "under test". Also, notice that the entire interaction is stubbed using hoverfly so in fact it does not even reach the actual GitHub server. I think we should simply bring it back and perhaps comment that it's stubbed repo as it's clearly not obvious based on what happened here :)

bartoszmajsak (changes requested):
Thanks for taking care of it @mabartos! I think we have to bring back one test which was deleted by "accident" and we should be good.

drone-webdriver/src/test/java/org/jboss/arquillian/drone/webdriver/binary/downloading/source/GitHubSourceLatestReleaseFromTagsTestCase.java
mabartos: Aaah, you're definitely right :) I'll do it

mabartos:
@bartoszmajsak I couldn't execute the `GithubSourceLatest*` tests on my local machine.

mabartos:
@bartoszmajsak Can you please review the PR again? Thanks

mabartos:
The `GithubSourceLatest*` tests passed.

PR #615, opened 2025-10-29
CI: add browser to the test matrix

rhusar:
This was missing entirely in the GHA CI setup and done on CircleCI only without platform matrix.


rhusar:
@lprimak Anything to add?

